PR #1, opened 2017-10-11
Feature/effect

ShMcK:
Adding effects.

Note: had to comment out some view logic

// init.js
```js
if (config.view) {
    registerViewImplementation(config.view)
  }
```

&

// model.js
```js
  // NOTE: not sure about this
  // NOTE: should only be called if a view is used
  // createViews(model)
```

Also commented out the view tests.


src/model.js
```diff
@@ -14,6 +15,9 @@ export default (model: $model): void => {
   createReducersAndUpdateStore(model)
 
   createDispatchers(model)
+  createEffects(model)
 
-  createViews(model)
+  // NOTE: not sure about this
```
blairbodnar: Yep yep, agreed. Can talk about this :)

PR #2, opened 2017-10-11
Feature/examples

ShMcK:
Adds local copies of examples & packages. These are essentially cloned from the existing repos @blairbodnar worked on. Packages reference each other using "file:../..", for easier package development and testing.

It's certainly up for discussion, but I think it may be easier to put everything in one place. This is how projects such as [dva](https://github.com/dvajs/dva) have now moved towards. The problem with having a lot of repos is the same as having a lot of folders. Its less scalable if the packages are connected with each other.

In addition, we can run all of the tests on all of the packages as well to make sure everything stays in sync.


blairbodnar:
Yep yep, totally cool with this. I was debating going with this approach as well!

Issue #6, opened 2017-10-12
Remove the API "shortcuts"


blairbodnar:
Remove the shortcuts. Eg.
export {
  init,
  init as i,
  model,
  model as m,
  ...
}

PR #22, opened 2017-10-13
Fixes #17 Expose store in public API

### LINKED ISSUE #17: Expose store in public API

blairbodnar:
Fixes #17 

@ShMcK 

Ran into some issues when trying to expose the store reference directly. 

The store was always resolving to null. This didn't work:
```js
// src/index.js
import { store } from './src/store'
export default {
  store
}
```

So, I exposed it via a function: `getStore`. So the following works, all tests pass :)
```js
// src/index.js
import { getStore } from './src/store'
export default {
  getStore
}

// elsewhere
import { getStore } from './src/index'
const store = getStore()
```

### DIFF
--- a/src/dispatch.js
+++ b/src/dispatch.js
@@ -3,7 +3,7 @@
 /**
  * dispatch
  */
-import { store } from './store'
+import { getStore } from './store'
 
 export let dispatch = {} // eslint-disable-line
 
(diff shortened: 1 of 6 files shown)


blairbodnar:
And when I say all tests pass, I mean all except the effects one that is still broken - unrelated to this.  :)

ShMcK:
`getStore` makes a lot of sense. 

As long as `dispatch` is provided, users likely won't need other access to `store`.

blairbodnar:
Oops, forgot to note that this also fixes #6 Remove the API "shortcuts"

Issue #11, opened 2017-10-12
subscriptions


ShMcK:
Look into a hooks API for models.

ShMcK:
Consider the following models as an example:

```js
{
 name: 'count',
 state: 0,
 reducers: {
    addOne: (state) => state + 1,
    addBy: (state, payload) => state + payload,
  },
  /* ...hooks */
}

{
  name: 'form',
  state: {},
  reducers: {
    submit: () => {}
  }
}
```

---

### Hooks API Specs

A hooks API should fit all possible combinations of things that someone might want to happen:

1. `form/submit` triggers `addOne`, but only the first time, then the subscriber disappears. In "redux-saga" this is called `take`
2. `form/submit` triggers `addOne` every time it is fired. In "redux-saga" this is called `takeEvery`
3. `form/submit` triggers `addBy` with the payload 5
4. `form/submit` triggers both `addOne` and `addBy`, one as a "take" and the other as a "takeEvery"
5. `form/submit` triggers `addOne`, but only if `addBy` is not triggered, as `addBy` is given precendence

7. A user should be able to pattern match a hook.

```js
take('*/submit', addOne())
take('form/*', addOne())
```

In the example above, all `submit` actions would be taken.

8. A hook should not need to trigger an action in another model

---

There may be other scenarios. @blairbodnar, let me know if you can think of any.

ShMcK:
At its simplest form there are two parts. I'll call them an "Action Matcher" and "On Action" for now. But those definitely aren't good names.

### Action Matcher
The action you're listening for. 

Examples: `form/submit`, `*/submit`, `form/*

### On Action
The function called when the action is matched.

This could probably be restricted to `effects` or `reducers` only. I'm not sure at this point.

It should also consider whether an action is listened for like a `take` or `takeEvery`

ShMcK:
My thoughts so far....

```js
{
  hooks: {
    'model/actionName': /* something here */
    '*/actionName': /* something here */
  }
}
```

ShMcK:
A possibility:

```js
{
  hooks: {
    'model/actionName': () => take(this.AddOne()),
    '*/actionName': (action) => takeEvery(this.addBy(action.payload))
  }
}
```

ShMcK:
How matchers might work:

1) look for an exact match
2) look for a pattern match

Matchers would have to be performant to make it work well. Pattern matching a lot of hooks may create performance issues on every action triggered.

PR #34, opened 2017-10-15
Feature/hooks

ShMcK:
Implements action type matching hooks & pattern matching hooks.

Progress on #11. I've updated the notes with checkboxes/checkmarks to show what is completed.

See `hooks.test.js` for examples.

##### Notes:

Pattern matching action types is handled separately. There will be some performance concerns if a lot of pattern matching hooks are added, devs should be warned against using them.

##### Remaining Todos:

- [ ] consider binding the context of a hook to its model. 
- [ ] implement a way for users to dynamically add or remove hooks. The code is there, but it is not exposed. Maybe `hooks.remove(name)`, `hooks.add(name)` ?
- [ ] implement take once only hooks (currently they take every time)

### DIFF
--- a/flow-typed/local/model.js
+++ b/flow-typed/local/model.js
@@ -10,7 +10,7 @@ declare type $model = {
  selectors?: {
   [name: string]: (state: any, arg?: any) => any,
  },
- hook?: {
-  [actionMatch: string]: (action: Object) => void,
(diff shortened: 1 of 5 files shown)


blairbodnar:
Awesome job on this @ShMcK , thanks!

Issue #7, opened 2017-10-12
look into a way to reference "this" within a model


ShMcK:
When referencing state within an effect, it would be nice if we could reference state with something like:

```js
reducers: {
  total: 0
},
effects: {
  add: (payload) => this.total + payload
}
```

Is this an idea worth considering?

ShMcK:
See example referenced in #19.

ShMcK:
Some possibilities:

1. Create with objects, but then transform them into classes
2. Use `call` or `bind` to give the model context

ShMcK:
In this case, `this` should actually be `dispatch[modelName]`

blairbodnar:
If in dev mode, after init, run each effect and hook as a string and pattern match for external dispatch calls. Trigger console.warn with a warning message / suggestion if there is a match.

ShMcK:
I currently have an implementation that works in the following scenarios:

1. function scope - PASS
```js
effects: {
  asyncAddOne: async function() {
    await this.addOne()
  }
}
```

2. object function shorthand - PASS
```js
effects: {
  asyncAddOne() {
    this.addOne()
  }
}
```

3. arrow function - FAIL
```js
effects: {
  asyncAddOne: async () => {
    console.log(this) // undefined
    await this.addOne()
  }
}
```

Arrow function will "always" reference its scope, and cannot be `bind`ed to another scope.

ShMcK:
There may be to test if a function is an arrow function, then override it with a regular function.

ShMcK:
It seems the most reliable way to detect an arrow function is by calling `.toString()` on a function, and making a regex. I couldn't find an available regex. But working on one now. Will update this as I go. 

```js
^((async)\s+)?((\(([a-zA-Z0-9_-]+?)?\))|[a-zA-Z0-9_-]+)?(\s?)+=>
```

- [x] () => anything
- [x] (something) => anything
- [ ] (athing, bthing) => something
- [x] async () => something
- [x] async (something) => anything
- [ ] async (athing, bthing) => anything
- [x] ()=>something
- [x] athing=>athing
- [x] athing => athing
- [x] async    () => {}

ShMcK:
Regex to check for arrow functions.
```js
const isArrowFunction = (fn) => {
  const isArrowFunctionRegex = /^((async)\s+)?((\(([a-zA-Z0-9,\s_-]+?)?\))|[a-zA-Z0-9_-]+)?(\s?)+=>/
[...]

PR #39, opened 2017-10-16
add `this` context to effects

ShMcK:
Progress with #7 .

Allows the following functionality:

```js
{
  name: 'example',
  state: 0,
  reducers: {
    addOne: (state) => state + 1
  },
  effects: {
    callAddOne() {
      // this references the local model dispatch
      this.addOne()
    }
  }
}
```

Currently works with.
- [x] - object function shorthand
- [x] - function
- [ ] - fat arrow functions

### DIFF
--- a/src/effects.js
+++ b/src/effects.js
@@ -8,8 +8,11 @@ export const effects = {}
  */
 export const createEffects = (model: $model) => {
   Object.keys(model.effects || {}).forEach((effectName: string) => {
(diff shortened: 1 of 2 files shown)


ShMcK:
Currently does not work with arrow functions. There is no clear way to detect arrow functions, and they are often transpiled by a pre-compiler, making it likely impossible. See #7 for efforts.

PR #41, opened 2017-10-16
WIP: Feature/plugins

### LINKED ISSUE #8: plugins API ?
We should create a plugins API under init. This can allow for merging store middleware, setting up models, etc.

Plugins may be:
A) an array of objects

```js
init({
  plugins: [plugin(), plugin2()]
})
```

B) an array of strings that internally references an existing local package

```js
init({
  plugins: ['plugin1', 'plugin2']
})
```

C) Another possible API to consider, as used by DVA:

```js
const app = init()
// sets up plugin
app.use(plugin())
```

Regardless, we should decide on a plugin API with some reasoning.

Some things to consider:

1) The plugins are more efficient if they run on startup, which leads me towards A & B.
2) If we are to hold all packages locally, I'd lean towards B. If users are able to create their own packages, or add config settings to a package I'd go with A.

Personally, I'm leaning towards A. What are your thoughts?

ShMcK:
@blairbodnar, brilliant work so far.

Resolves #8.

##### PR is blocked from merging. 
I added the "Work in Progress" bot to Github, which will block any PRs with WIP in the title from merging. So don't worry, this is safe! Try to merge #43, not gonna happen (unless you change the title).

We can use this PR stream to have a dialogue about the changes.


ShMcK:
Note: for tests, you can simply comment them out by adding an x in front. `xdescribe` or `describe.skip()` both work. See [Jest docs](http://facebook.github.io/jest/docs/en/api.html#describeskipname-fn).

ShMcK:
Is the third `onModel` param necessary? ie. `(dispatch)`. If dispatch is a plugin, can it not be accessed from `pluginExports.dispatch`?

ShMcK:
### TODOs
- [ ] validate plugins
- [ ] flattened exports
- [ ] bring back tests & test coverage

ShMcK:
`onInit`: if there are not any params getting passed in, why not just set it to an array?

blairbodnar:
Awesome, thanks @ShMcK . And excellent work with the hooks! 🔥🔥🔥

> You can skip tests simply comment them out by adding an x in front. xdescribe or describe.skip() both work. See Jest docs.
- Awesome! Thanks. Will do going forward 👍

> onInit: if there are not any params getting passed in, why not just set it to an array?
- Yep, makes sense! - Let's change it. I'll add it to the todos you listed

> Is the third onModel param necessary? ie. (dispatch). If dispatch is a plugin, can it not be accessed from pluginExports.dispatch?
- Maybe... However, we might need it to create the dispatch plugin itself?
[https://github.com/rematch/rematch/blob/feature/plugins/src/plugins/dispatch.js#L12](https://github.com/rematch/rematch/blob/feature/plugins/src/plugins/dispatch.js#L12)

ShMcK:
@blairbodnar, one quick check:

1. Dispatch is created using `store.dispatch`
2. Then "effects" use `store.dispatch` or the new custom `dispatch`?

If its possible for every plugin to rely on `store.dispatch`, then the plugins won't be bound to each other, and can possibly even be added in any order.

ShMcK:
Changes:

- I added tests back
- changed onInit to be a key/val object
- restructure/reorganized files
- added tests back
- broke the app in some small undiscoverable way that will drive you crazy

blairbodnar:
> If its possible for every plugin to rely on store.dispatch, then the plugins won't be bound to each other, and can possibly even be added in any order.
* I think so, Looks like effects has the normal store.dispatch passed in, so looks ok?

> broke the app in some small undiscoverable way that will drive you crazy
* haha, fun times! no worries 😝

blairbodnar:
@ShMcK Fixed! 😎

ShMcK:
Great.

### Questions

1. If a plugin takes the param of "pluginExports", does it also need `onModel` "exports" as a param?
2. Would the plugin pipeline be better suited as a pipeline that returns a value that is passed to the next pipe? In the same way that promises work. Currently we have a pipeline that mutates a set of variables. I'm not sure where I stand on this, or how a pure functional pipeline might work, but I suppose it may look nicer if we could chain everything. I wouldn't approach trying this without full test coverage first.

### TODOs

##### Architecture
- [ ] export dispatch
- [ ] find a way with the new architecture to make direct dispatch possible
- [ ] export select
- [ ] hide unnecessary exports
- [ ] validate plugins

##### Tests
- [ ] get uncommented tests passing again
- [ ] add select tests that test the use of select within a setup app
- [ ] write a test method for testing plugins

blairbodnar:
Great thoughts and notes @ShMcK , thanks!
Our brains are on the same

> If a plugin takes the param of "pluginExports", does it also need onModel "exports" as a param?
* Nope, I don't think it does!

> Would the plugin pipeline be better suited as a pipeline that returns a value that is passed to the next pipe? In the same way that promises work. Currently we have a pipeline that mutates a set of variables. I'm not sure where I stand on this, or how a pure functional pipeline might work, but I suppose it may look nicer if we could chain everything. I wouldn't approach trying this without full test coverage first.
* It's funny, I was thinking the exact same thing as I was coding this up and getting it to work. My gut started feeling a little queezy with all the mutating going on. I think it would kick ass if we were able to have this work in more of a functional pipeline. Right now it feels like a little too much like magic 😜.

> onInit no longer looks like a great name. Any ideas?
* I agree. Maybe `setup`?

### Further thinking...
#### Currently, we have this...
(I'll refer to it as `onInit` in the following example so it looks familiar to what we've seen already)
```js
// dispatch
onInit: { dispatch: {} } 

// effects
onInit: { effects: {} }

// selectors
onInit: { select: {} }

// hooks
onInit: {
  hooks: new Map(),
  patternHooks: new Map()
},
```

#### What if `onInit` told rematch what needs to be exposed to the public API, and what needs to be exposed to the internal API. This could help in a better pipeline design. Something along the lines of this maybe?
```js
// dispatch
onInit: {
  publicAPI: { dispatch: {} },
  internalAPI: null
}

// effects
onInit: {
  publicAPI: { effects: {} },
  internalAPI: null
}

// selectors
onInit: {
  publicAPI: { select: {} },
  internalAPI: null
}

// hooks 
onInit: {
  publicAPI: null,
  internalAPI: {
    hooks: new Map(),
    patternHooks: new Map()
},

```

ShMcK:
Differentiating between exports & internal API is a good idea.

It may be easier just to flatten it, and skip `onInit` altogether. Something like this:

```js
{
  exportsAPI: { dispatch: {} },
  internalAPI: null
}
```

I'm also not yet sure if we will be able to get external exports working within this pipeline. I believe it will require a read only export placeholder.

```js
export const dispatch = {}
export const select = {}
```

If we can solve that problem, we can come back to the plugins API.

ShMcK:
pushed a version that no longer requires "onInit", but rather handles it all in basic js. 

Also revives test coverage for "dispatch" & "effects". Will work on reviving the remaining tests.

ShMcK:
Tests remaining
- [ ] select test rewrite
- [ ] hooks: pattern matching
- [ ] effects - dispatch with local this
- [ ] plugins - write tests for these once the API is sorted out

blairbodnar:
@ShMcK Nice work!

I have an idea for named exports... gonna take a stab at it.

blairbodnar:
@ShMcK Nice job - You basically implemented the idea I had for named exports! Just having the plugin export out anything it wants to make available for the consumer. It's nice and simple. I like it!

(Sure, the named exports aren't part of the core rematch package, but I think that's totally fine.)

### [show-me-the-models-plugin.js](https://www.youtube.com/watch?v=FCgy0dphOGc)
```js
export const models = {}
export default {
  onModel: (model) => {
    models[model.name] = model
  }
}
```
### app.js
```js
import { init, model } from 'rematch-x'
import ShowMeTheModelsPlugin, { models } from 'show-me-the-models-plugin'

init({
  plugins: [ShowMeTheModelsPlugin]
})

model({
  name: 'count',
  state: 0
})

model({
  name: 'todos',
  state: []
})

// Show me the models!
console.log(models)
```

### somewhere-else.js
```js
import { models } from 'show-me-the-models-plugin'

// Show me the models!
console.log(models)
```

ShMcK:
That's true. Exports can live within the plugin, issue solved!

Issue #51, opened 2017-10-19
Unsure why test fails


ShMcK:
I'm not sure why the following plugins test is failing in "plugins.test.js".

```js
 xtest('should add middleware hooks', () => {
    const m1 = () => next => action => next(action)
    const m2 = () => next => action => next(action)
    const hooks = [{
      middleware: m1,
    }, {
      middleware: m2
    }]
    createPlugins(hooks, [])
    expect(pluginMiddlwares).toEqual([m1, m2])
  })
```

ShMcK:
closed once #56 is ready to merge.

PR #56, opened 2017-10-22
Subscription improvements

### LINKED ISSUE #52: Test hooks with multiple listeners
Possible bug:

I believe in the current form, new hooks may overwrite existing hooks.

##### TODO

- [ ] write tests with multiple hooks of the same name

ShMcK:
Closes #52. Once resolved, should handle multiple subscriptions with the same name in different models.

### Remaining Issue 

An action is triggered twice within middleware.
- 1st) a dispatched action is sent
- 2nd) the dispatch function, without the dispatch is sent

This is very strange.

The result is that a subscription is fired twice. However, after looking into it, I don't believe the fault is with the subscriptions code. If I comment each part out individually, I still have the same issue. I believe we may need more tests for redux store middleware.

See the failing test in "subscriptions.test.js". Also, in order to see what was described above, go into "subscriptions.js" and add a `console.log(next, action)` before `next(action)` is returned. Notice that next is a dispatched function, but then the same function without a dispatch the second time.

I'll leave this PR as WIP until it is resolved.


ShMcK:
This may also be related to #51. I believe there is an issue that is only created when we add multiple middleware.

ShMcK:
There's a hint as to what the multiple dispatch problem is.

For every model created with a subscription, there is an additional dispatch called.

1 model with hooks: dispatch called twice
2 models with hooks: dispatch called three times

ShMcK:
Issue resolved by loading the imports within the test block. See [resetModules](http://facebook.github.io/jest/docs/en/jest-object.html#jestresetmodules) docs.

```js
beforeEach(() => {
  jest.resetModules()
})

test('', () => {
  const { someImport } = require('../src')
})
```

Keep this mind, it'll probably crop up again.

ShMcK:
Note that some pattern matched subscriptions are not possible:

"*" for example will result in an infinite loop. 

There should probably be a way to ensure that users can't create subscriptions that match any of their reducers or effects. I'll make a note.

ShMcK:
Tried working on another branch, but I see some merge conflicts coming. going to merge this in and build on top.

PR #60, opened 2017-10-23
closes #50. Autobind selector to model state as first param

### LINKED ISSUE #50: selectors auto bind to getState()
Should selectors automatically be passed `getState()` as the first parameter?

##### Current

```js
state = 1
select.count.double(state) = 2
```

##### Proposed
```js
state = 1
select.count.double() = 2
```

---

##### Current

```js
state = { '2': { name: 'Joe' } }
select.user.getById(state, '2') = { name: 'Joe' }
```

##### Proposed
```js
state = { '2': { name: 'Joe' } }
select.user.getById('2') = { name: 'Joe' }
```

ShMcK:
Selectors no longer needs state as the first param.

```js
// before:
select.a.b(state)

// after PR:
select.a.b() // automatically passes store in
```

See examples in "select.test.js"

### DIFF
--- a/examples/react-redux-count/src/index.js
+++ b/examples/react-redux-count/src/index.js
@@ -67,7 +67,7 @@ const App = ({ valueA, valueB, valueADoubled, asyncAIncr, incrB, incrA }) => (
 const AppContainer = connect(state => ({
   valueA: state.countA,
   valueB: state.countB,
-  valueADoubled : select.countA.double(state),
+  valueADoubled : select.countA.double(),
   incrA: () => dispatch.countA.increment(),
   asyncAIncr: () => dispatch.countA.asyncIncrement(),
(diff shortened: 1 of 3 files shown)


blairbodnar:
Very nice! 

Just two thoughts that I think are important to keep in mind that selectors should be able to do:

1. Integrate easily with the view layers. Eg react-redux's `connect`
2. Integrate easily with [reslect](https://github.com/reactjs/reselect) / or other similar techniques to make them more efficient (and maybe composable)

Maybe we can merge this PR and make a new issue with the above two points to ensure we keep them on the roadmap?

ShMcK:
I think we should demonstrate a basic demo using "reselect" before merging this PR. I haven't tried using it together with auto-bound state.

ShMcK:
I'll make an issue about selectors.

PR #61, opened 2017-10-23
closes #47. Add model option to plugin

### LINKED ISSUE #47: add "model" to plugin creator
It should be possible to create a plugin that adds a model. For example "loading".

ShMcK:
Now add a model to a plugin.

Question though: Would a plugin ever require more than one model? This current PR only allows a single model. This opens the question.... could your entire app be broken into a series of plugins?

You right now.


blairbodnar:
lol.

Good question... It could be possible I think! haha

Just gonna let my brain digest it.

PR #82, opened 2017-10-28
WIP - Plugin/loading

ShMcK:
### Todos
- [ ] make it work

### How it works
1) create a plugin with config
2) plugin creates a "loading" model
3) plugin maps over all dispatches, and if a dispatch is an effect, it modifies the dispatch in the following way:

```js
dispatch[namespace][action] = async function (props) {
  dispatch.loading.show({ namespace, action })
  await fn(props) // waits for the original dispatch to finish
  dispatch.loading.hide({ namespace, action })
}
```


blairbodnar:
I also accidentally remote delete this from webstorm. ughhhhhh. Sorry man! not sure if you still want this one open.

ShMcK:
close in favor of #85

PR #84, opened 2017-10-29
WIP - plugins api refactor

blairbodnar:
@ShMcK 

I'm typing this in a rush as I have to head to a dinner! Just playing around with some different ideas.

I'm not sure if I'm quite done with this, but as it stands:
- All tests pass.
- I changed all the core plugins to work with this approach. 

Take a look when you have a sec.

Maybe it makes life easier if you try and make some plugins with this API? 

#### plugins:
```js
export default {
  
  /*
    Expose stuff to other plugins.
    For other plugins: available as first param in 'internalInit'
    For the 'consumer': Available as normal. Just import it. see src/index.js for an example
  */
  expose: { select: {} },
  
  internalInit: (exposed) => ({
   onModel(model: $model) {
      exposed.select[model.name] = {}
      // do stuff with exposed things from other plugins
    }
  })
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "build": "webpack --env build",
     "dev": "webpack --progress --colors --watch --env dev",
     "lint": "eslint ./src",
-    "test": "jest"
+    "test": "jest --watch"
   },
   "dependencies": {
     "redux": "^3.7.2"
--- a/src/core.js
+++ b/src/core.js
@@ -5,6 +5,7 @@ import createModel from './model'
 
 export const modelHooks = []
 export const pluginMiddlewares = []
+const exposed = {}
 
 const validatePlugin = (plugin: $plugin) => validate([
   [
@@ -17,14 +18,19 @@ const validatePlugin = (plugin: $plugin) => validate([
   ],
 ])
 
-const buildPlugin = (createPlugin): $plugin => {
-  const { dispatch } = getStore()
-  return createPlugin(dispatch)
+export const populateExpose = (plugins) => {
+  plugins.forEach(p => {
+    if (p.expose) {
+      Object.keys(p.expose).forEach(key => {
+        exposed[key] = p.expose[key]
+      })
+    }
+  })
 }
 
 export const addPluginMiddleware = (plugins) => {
   plugins.forEach(createPlugin => {
-    const plugin = createPlugin()
+    const plugin = createPlugin.internalInit(exposed)
     if (plugin.middleware) {
       pluginMiddlewares.push(plugin.middleware)
     }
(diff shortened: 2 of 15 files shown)


ShMcK:
I see. This is similar to your first approach, before I tore it apart and used file scoping of exports.

I think this solution is much cleaner for a few reasons.
- everything in one place
- access to all core plugins without exports

### Suggestions

I think it would look a bit nicer though, if exposed were bound to the plugin.

```js
export const internalInit = (exposed) => ({
  exposed: {
    select: {}
  },
  onModel(model: $model) {
    exposed.select[model.name] = {}
    Object.keys(model.selectors || {}).forEach((selectorName: string) => {
      exposed.select[model.name][selectorName] = (state: any, ...args) =>
        model.selectors[selectorName](state[model.name], ...args)
    })
  }
})
```

blairbodnar:
@ShMcK 

Thanks, glad you like it!

I worked some more on it tonight, and I pushed to this branch. Hopefully I didn't mess with anything you were doing? I didn't see any pushes to it from you so I figured it was ok push. I probably should have done another branch off of this? Sorry If I'm messing this up! Not sure if I should have done a branch of a branch?

Anyways,

* I like the suggestion you mentioned above. I took a stab at implementing it that way but had no luck getting something to work. So, I left it as is for now.
* Got the loading plugin working. Wrote some tests. Should probably be more, but I think it works ok!
* All tests passing

Todo
- [ ] Implement a plugin config solution.
- [ ] Make code nice. Flow. eslint. Better naming? Better comments. 
- [ ] Sleep 😴 

---
For plugin configs, I tried this approach but had no luck getting it to work. It could be just because my brain is dead right now though...
```js
export default (config) => ({
  expose: { ... },
  init: { ... }
})
```

ShMcK:
I played around with some changes, but ran into a wall due to undefined imports. Although I did stumble into some optimizations, still had tests failing.

I was hoping to get plugins to be default function exports.

```js
export default (exposed) => ({
  expose: { key: value }
})
```

However, it seems there are only two options:


##### 1. Plugins as Objects
```js
export default {
  expose: { key: value },
  plugin: (exposed) => ({ ... })
}
```

Pros: allows shared internal access to exposed values
Cons: appears overly complicated

##### 2. Plugins as Functions with Exported Values
```js
export const key = value
export default (exposed) => ({ ... })
```

Pros: looks like regular javascript
Cons: plugins must maintain import paths to each other. This can get pretty fragile with new plugins

blairbodnar:
Alright, passing off to you for now @ShMcK 

In the last commit, I changed the loading plugin to a function that take a config as an arg. This seems to work ok because it isn't a core plugin. I ran into a wall when trying to convert the core plugins over because of the named exports. 

I'm off for now, go ahead and work your magic! 😎

PR #85, opened 2017-10-29
Feature/plugins api refactor

blairbodnar:
WTF. Ok I accidentally deleted this remote from webstorm. I need to step away from the computer now. Sorry! Ugh!


blairbodnar:
See the thread on #84 for the comments

Issue #72, opened 2017-10-27
Persist Plugin


ShMcK:
Create a persist plugin using redux-persist.

Issue #70, opened 2017-10-26
Root Reducer Enhancer


ShMcK:
Add a root reducer enhancer for implementing the new "redux-persist" 5.0.

Currently there doesn't seem to be a way to do this in DVA, so I'm thinking of moving over soon.

PR #88, opened 2017-10-30
Plugin Config & Persist Plugin

ShMcK:
Requires several changes to core

1. give plugins a config that can merge into the existing config

For example: a plugin can have a config with "extraReducers" or "extraEnhancers" that get merged into the current config.

This is complicated due to the fact that merging configs isn't as simple as a deep merge. It should probably work as follows:

- array
  - if already exists in the array, skip
  - add to the end of the array
- object
  - deep merge
- function
  - wrap the function with the next function 

2. `rootReducerEnhancer`, an optional wrapper around combine reducers

### TODO
- [x] allow a plugin config
- [x] complete config merge
  - [x] array
  - [x] object
  - [x] function
- [x] complete persist plugin
- [x] create `rootReducerEnhancer`


ShMcK:
Note:

The following test (plugin.test.js, 'should merge plugin configs into configs') triggers the following error:

```
 console.error node_modules/redux/lib/utils/warning.js:14
      Store does not have a valid reducer. Make sure the argument passed to combineReducers is an object whose values are reducers.
```

Should look into this at some point.

ShMcK:
Tests fail at combineReducers.

Issue is that current tests default the rootReducer to `state => state`. This lets our simple tests pass, like count, where state = 0. But combineReducers requires a root reducer with an object based state. Will try to resolve this.

```
 TypeError: Cannot convert undefined or null to object
        at Function.keys (<anonymous>)
```

ShMcK:
Tests passing!

But a few todos:

- [ ] Warnings in the console whenever a `customCombineReducers` is used.

```
   console.error node_modules/redux/lib/utils/warning.js:14
      Store does not have a valid reducer. Make sure the argument passed to combineReducers is an object whose values are reducers.
```

- [ ] Could use tests for purging state

ShMcK:
closes #70. closes #72 .

ShMcK:
Travis wasn't installing the deps within the plugins directory. 

Should work now. If you make a plugin, just add to the travis script.
If the travis script gets a bit long, we can make a script.

PR #91, opened 2017-10-31
Pre publish

ShMcK:
### TODID
Added some publishing tools:

1) moved CI installation scripts into "scripts/install.sh"
2) created a build script in "scripts/build.sh"
 - builds core lib
 - builds plugins

Setup npm packages for publishing.
 - npm ignore
 - filled out package.json's
 - added webpack configs for plugins

### TODO
- [ ] select a core package name

My suggestion: "@rematch/core". This is similar to how DVA does it.


ShMcK:
Packages can then be released by using [release-it](https://github.com/webpro/release-it)

Run the following to see what it would look like. With the command "dry run", it won't actually run the commands.

```
npm i -g release-it
release-it --dry-run --pre-release=alpha
```

Issue #97, opened 2017-11-03
@rematch/loading 0.1.0 - ReferenceError: regeneratorRuntime is not defined


blairbodnar:
This following code generates the error: "ReferenceError: regeneratorRuntime is not defined"
```js
import RematchLoading from '@rematch/loading'

init({
  plugins: [RematchLoading()]
})s
```

blairbodnar:
Didn't get a chance to look into this yet, but I'm wondering if it might be the same issue that we fixed with publishing `@rematch/core`? ([The babel fix we made going from 0.1.0-alpha.1 to 0.1.0-alpha.2](https://github.com/rematch/rematch/commit/433c90d9f059105618b78527d379012ffb7fcfd0)) 

Also, I haven't tried `@rematch/persist` yet... I'm gonna guess that this issue might be there as well.

ShMcK:
I think you're right.  I have a quick fix...

ShMcK:
I'll push a fix, but it may be worth looking into #93.

ShMcK:
Ah, it seems you're in an environment that is not supporting Async/Await.

You may need to use babel, typescript, or a polyfill. Learn more [from this babel issue](https://github.com/babel/babel/issues/5085)

suazithustra:
Would it be counterintuitive to just use fetch.then.catch in the effects instead?

suazithustra:
Babel is just fine, thanks.
```
{
  "presets": [
    ["@babel/preset-env", { "targets": { "node": "10" } }]
  ]
}
```

PR #100, opened 2017-11-03
add babelrc to each plugin with devDependencies

ShMcK:


ShMcK:
Just published. @blairbodnar can you check if this closes #97?

PR #103, opened 2017-11-04
README: add comparison of redux & rematch

ShMcK:
Adds a table comparing Redux & Rematch. I thought this was the most persuasive part of our presentation, so I stole it from your notes.

NOTE: The formatting may be messed up, I'm not sure if it will display correctly on Github. It's written in raw HTML in order to make a table with code blocks work. May need some editing as well.

Should look something like this, but hopefully with github styles:

screen shot 2017-11-04 at 9 26 44 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,26 +1,118 @@
+<style>
+  .badges > * {
+    margin: 0 0.5rem;
+  }
+</style>
+
 # Rematch
 
-<p style='display:flex;'>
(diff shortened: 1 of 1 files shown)


ShMcK:
it may be better to put the examples one after each other, in case people look at it on small screens. Wonder what you think?

ShMcK:
I think I'll merge it, but may revert it if the formatting is messed up.

ShMcK:
Formatting is messed up and it doesn't get syntax highlighting.

PR #105, opened 2017-11-04
add comparison of redux to readme

ShMcK:
Much happier with this version.

Let me know what you think.


ShMcK:
Added a "store" boilerplate bit as well...

PR #108, opened 2017-11-04
update Loading plugin readme, add config

ShMcK:


ShMcK:
Part of this is broken, trying to fix "effects".

PR #126, opened 2017-11-10
fix redux devtools.

### LINKED ISSUE #124: Issue with devtools
Debugger not working with Rematch.

See possible [fix](https://github.com/dvajs/dva/blob/aef20ab096a5ab3bd3209c8ae45a6f8f01a8c09e/packages/dva-core/src/createStore.js)

ShMcK:
closes #124 . Fixes broken redux devtools

Also allows users to pass in their own devtools config.

```js
init({
  devtoolsConfig: {},
})
```


ShMcK:
This was the most annoying problem with rematch. Happy again!

PR #135, opened 2017-11-13
Update purpose.md

blairbodnar:
Isn't store.dispatch() in vanilla redux a "global dispatch"?


ShMcK:
Oops. Actually, its only "rematch" for direct dispatch.

blairbodnar:
How about calling them "generated dispatchers"?

ShMcK:
I'm not sure if people will understand what we mean, unless there is a larger explanation.

blairbodnar:
Yeah, let's just scrap it for now?

Issue #76, opened 2017-10-28
Plugins vs. Packages


ShMcK:
I wondered if calling everything a "package" or a "plugin" makes sense.

### Plugin
Runs as a plugin in init
Example: Loading, reactRouter4, etc.

### Package
Works with rematch, but doesn't run as a plugin
Example: ReactRematch

Should we differentiate between the two? If so, how?

blairbodnar:
Hmm. I'm ok with "plugin" for now I think.

As for the package... I'm wondering if we even need them? I think They might just be recipes or examples? I'm thinking we probably don't need ReactRematch anymore... I think for using rematch with React it should just be an example or recipe showing how to use rematch with `react-redux`. 

Thoughts?

ShMcK:
I think having no "RematchReact" is definitely the simplest solution. 

Lets call it plugins then.

I do think we should preserve the existing rematch-react somewhere, in case we take up the cause again. Though it should be preserved in the git history if we ever come back to it.

ShMcK:
Some more options arise now:

##### should our existing plugins move into plugins?
This may cause issues with the webpack build.

##### should new plugins be npm packages?
I'm thinking so, as they may have their own dependencies.

PR #141, opened 2017-11-14
Packages/react

blairbodnar:
Phew. Sorry for the lack of documentation on this. I'm kinda going backwards, writing it from the integration side first, since I wasn't really sure the API I wanted.  I wanted to play with it as I wrote it and try some different ideas. 

I'm pretty happy with this! Still need to test performance and stuff like that, but I think it's looking pretty good!

```js
// when you have a model like this in Rematch...
const model = {
  name: 'count',
  state: 0,
  reducers: {
    up: () => {},
  },
  effects: {
    upAsync: () => {},
  },
  selectors: {
    doubled: () => {}
  }
}

// you can do this...
connect(['count'], ({ count }) => {
  // you have access to:
  
  // count.state // 0
  // count.dispatch.up
  // count.dispatch.upAsync
  // count.select.doubled() // call it just like that! no more passing in the state! :)

  // No need for accessing dispatch or select from global Rematch!
  // You can do it all from here.
  // and if you need to have two or more models connected in one component, 
  // just add it to the array above and you've got it!
  // Like this: connect(['count', 'auth'], ({ count, auth }) => { ... })(Component)
  // This allows flexibility for building lots of
  // small components (say one component per model) or you can have 
  // bigger components that need to use multiple models. 


  return ({
    ...
  })
})(Component)

```

Anyways, [I have a working example](https://github.com/rematch/rematch/tree/packages/react/packages/react). Try it out and see what you think. 

Also, I put it in a packages folder following the pattern you did for plugins. Does it look ok to you?  We discussed this a bit in #76 and yeah it doesn't seem right to put it in plugins.

### TODO
- [ ] tests
- [x] simple example
- [ ] complex example (I'm confident it'll work nice - I have a CRUD todos app example almost done)
- [ ] eslint, flow, etc


ShMcK:
I see, the example is "model-centric", rather than type specific.

So rather than `dispatch.model.name`, its `model.dispatch.name`. Interesting idea, and might be something to it. After all, it makes sense than you're calling methods on the model. 

I have a few concerns:

- hard to convert from basic redux. In the previous way, rematch was limited to just the data, and everything works the same as before. It seems almost so different that it can be unattractive to people that already use react-redux.
- not sure about the duplication here `connect(['count'], ({ count }) => { ... }`. It doesn't seem simpler than connect.
- the majority of the time we are calling "state". I can see I would have to use a lot of "model.state.x" references, rather than just "state.x"

blairbodnar:
Thanks @ShMcK 

---
> - hard to convert from basic redux. In the previous way, rematch was limited to just the data, and everything works the same as before. It seems almost so different that it can be unattractive to people that already use react-redux.

True yeah. I guess it was written largely based on how I personally might like to work with rematch / mirror, etc - The nice thing is this 'package' doesn't even really have to be included with rematch at all. The src/index.js is just basically a kind of a recipe for working with react-redux in a certain way... This could just be a recipe for people who might want to work in a "model-centric"mindset.

---
> - not sure about the duplication here `connect(['count'], ({ count }) => { ... }`. It doesn't seem simpler than connect. Why not just `connect(({ count }) => { ... }`?

Yeah, could probably figure out a way to make that work

---
> - the majority of the time we are calling "state". I can see I would have to use a lot of "model.state.x" references, rather than just "state.x" or a destructured "{ x } "

I think it's about the same?
`state.count` vs `count.state`
`state.todos['1'].text` vs `todos.state['1'].text`

---
> - seems like performance may be hurt. 

- Maybe! Would be interesting to compare.

---
So yeah, all in all - Maybe scrap this and make it a recipe instead? Or not include it all?

ShMcK:
Its probably best to avoid too many options in how to do things. The kind of fragmentation can lead to problems down the line.

Maybe we should have a folder called "experimental" where we play with new features?

If someone sees "packages" they think its the one true way to do things, but if they see "experimental", then there's no issue.

blairbodnar:
Yeah, I agree - Don't wanna have too many options.

I'm cool with just closing the PR. No need for the experimental idea unless you really want it?

ShMcK:
I think its a cool idea. "experiments" or "experimental". I dunno, what do you think ?

blairbodnar:
Ok cool sure! Hmm, I kinda like "experiments". You?

ShMcK:
Works for me :)

Who knows what will come of it. We scientists now.

blairbodnar:
@ShMcK 

1. I updated my master.
2. I ran `git rebase master` my `Packages/react` branch
3. I made the necessary changes to move code from "packages" to "experiements"
4. I commited
5. I tried to push, and got errors saying I had to merge
6. So I merged...

Stuff got messed. :(

[moved from "packages" to "experiments"](https://github.com/rematch/rematch/pull/141/commits/cfd6cb97a77caafb91f3e61c1dd1fd5a61b8721b) is what I wanted to push.

blairbodnar:
No big deal if we just want to kill this branch/PR - I can do it again from a fresh start.

ShMcK:
Maybe a freshstart might be easiest. You can just admin merge in when you think its ready.

blairbodnar:
Let's kill this. I'll come back to it sometime!

PR #153, opened 2017-11-20
Plugin/updated

### LINKED ISSUE #77: Last Updated Plugin
Collect time stamps for the last time a set of data was updated

- call action when action triggering reducer is called
- access to list of actions (excluding effects)
- last updated model

ShMcK:
closes #77.


ShMcK:
Just realized my commit message there was "update updated docs". Hope that makes sense.

PR #166, opened 2017-11-26
migrate flow -> typescript, webpack -> rollup

ShMcK:
##### Pros
- improved type coverage
- smaller & more readable compiled output compared to Babel
- no more need for long lists of babel packages, especially "babel-runtime" & the one that adds "async/await"

##### Cons
- have to run "tsc --watch src" while developing, or setup a typescript watcher in your editor
- moves from eslint to tslint

##### Pros
- smaller package size (54kb -> 11kb, partially due to TS)
- builds for different environments (umd, es, cjs)
- simpler build config
- everybody else is using it, so maybe its good?

##### Cons
- zip zilch nada?

### Todo:
- [x] fix plugins
- [x] fix plugin tests
- [x] cleanup linter
- [x] add badge with package size to README
- [x] fix CI build
- [x] fix coverage issue from Typescript file polyfills


ShMcK:
Now the plugins are larger in size than the core, which includes "redux" as a dep.

Have to fix that.

ShMcK:
Just have to fix a few persist tests...

PR #171, opened 2017-12-01
Fixes

### LINKED ISSUE #170: Issue if payload is false
Should check if "payload" and "meta" are type undefined.

ShMcK:
Closes #170.


ShMcK:
@blairbodnar - Good catch on this bug :)

Issue #174, opened 2017-12-02
Test on Windows


ShMcK:
Nan also pointed out that our build script doesn't work on windows.

### Things that fail on Windows

- `&& NODE_ENV='production'` fails. Solution: use `cross-env` package in the build scripts.

- Npm file paths. `"@rematch/core": "file:../../.."`

It may also be worth testing more on Windows altogether for bugs.

Issue #168, opened 2017-11-29
Enhancers, Redux-Reset


ShMcK:
Add Redux-Reset to persist plugin for resetting state.
https://github.com/wwayne/redux-reset

Redux-Persist can only purge local storage.

---

Will also require a "redux" enhancers API.

```js
redux: {
  enhancers: [enhancer1]
}
```

PR #178, opened 2017-12-05
capture store from init, rather than getStore export

### LINKED ISSUE #172: Init creates store
Nan had a good idea.

We could eliminate `getStore` if `init` simply returns the store.

ShMcK:
closes #172.

### Breaking Changes

1. getGone, getStore

```
const store = init()
```

2. Plugin.onStoreCreated accepts store, rather getStore.

- [x] core
- [x] docs
- [x] plugins
- [x] tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,6 @@ The **model** brings together state, reducers, async actions & action creators i
 
 #### models.js
 ```js
-import { dispatch } from '@rematch/core'
-
 export const count = {
   state: 0, // initial state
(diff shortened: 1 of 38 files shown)


ShMcK:
Second issue closes one bug on windows. closes #174.

There may be others, probably worth testing.

ShMcK:
closes #168. 

Add potential "redux.enhancers" in config.

ShMcK:
Breaking changes... hope that's the last time I write that.

Issue #177, opened 2017-12-03
Middleware strangeness


ShMcK:
### Problem

I added some custom middleware to reset the state.

```js
import * as models from '../../models'

const initialState = Object.keys(models).reduce((state, name) => {
  state[name] = models[name].state
  return state
}, {})

export default {
  init: () => ({
    middleware: () => next => action => {
      console.log('reset state')
      return ((action.type === 'RESET') ? initialState : next(action))
    }
  })
}
```

Then dispatch `RESET`. 

### Expected

1. 'reset state' in log
2. See 'RESET' in Devtools
3. State set to initial states

### Result

1. 'reset state' in log
2. no action in Devtools
3. state unchanged

### Possible Causes

- There may be an interaction by plugins causing the problem.
- Middleware may be working incorrectly
- Order of middleware may be causing an issue

?

ShMcK:
Issue closes in favor of #177.

PR #181, opened 2017-12-09
redux.rootReducers -  a hook at reducers base that returns next state

### LINKED ISSUE #168: Enhancers, Redux-Reset

ShMcK:
References #177 & would close #168.

This PR so far only outlines the middleware with a failing test.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -3,6 +3,8 @@
 * [Introduction](README.md)
 * [Purpose](./docs/purpose.md)
 * [API Reference](./docs/api.md)
+  * [Redux API](./docs/reduxApi.md)
+  * [Plugins API](./docs/pluginsApi.md)
 * Recipes
   * [Devtools](./docs/recipes/devtools.md)
   * [React](./docs/recipes/react.md)
@@ -11,5 +13,4 @@
   * [Persist](./plugins/persist/README.md)
   * [Updated](./plugin/updated/README.md)
   * [React Navigation](./plugins/react-navigation/README.md)
-  * [Plugins API](./docs/pluginsApi.md)
 * [Inspiration](./docs/inspiration.md)
--- a/docs/api.md
+++ b/docs/api.md
@@ -10,7 +10,7 @@
   - [models](#models)
   - [initialState](#initialState)
   - [plugins](#plugins)
-  - [redux](./initReduxApi.md)
+  - [redux](#redux)
 
 
 ## dispatch
@@ -199,7 +199,7 @@ Plugins are custom sets of init configurations or internal hooks that can add fe
 
 Read more about existing [plugins](./plugins) or about how to create your own plugins using the [plugins API](./pluginsAPI).
 
-### Init Redux (continued)
+### redux
 
 There are situations where you might want to access Redux directly. You may want to:
 
@@ -207,4 +207,4 @@ There are situations where you might want to access Redux directly. You may want
 - add middleware
 - create a custom plugin
 
-For a complete summary of all redux options, see the [init Redux API](./initReduxApi.md).
+For a complete summary of all redux options, see the [init Redux API](./reduxApi.md).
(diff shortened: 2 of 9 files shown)


ShMcK:
After looking into this, I realize redux middleware is not meant to change state. It can listen to actions, and manipulate those actions, but changes can only be made through reducers.

### Not a bug

In order to return a different state, you need to overwrite the "root reducer".

```js
const rootReducer = (state, action) => {
  // manipulate state here
  return appReducer(state, action)
}
```

We could create a hook in here on startup, and give it a potential name like `redux.rootReducerMiddleware`?

If `redux.rootReducerMiddleware` then wrap something in a `createRootReducer(combineReducers(state, action))` function, else use `combineReducers`.

### Use cases

Adding a state `reset`, such as on logout, that simply sets state to `undefined`, thereby taking the default reducer states.

Adding more powerful middleware that can manipulate the state, such as an optimized loading indicator plugin.

Somehow this PR became an issue...

ShMcK:
### Proposal:

```ts
redux: {
   rootReducerWrapper?: (rootReducer: Reducer<any>) => (state: any, action: Action) => any,
}
```

### Implementation

```js
let rootReducer: Reducer<any> = mergeReducers()
if (redux.rootReducerWrapper) {
  rootReducer = redux.rootReducerWrapper(rootReducer)
}
```

Works as of last commit.

ShMcK:
Changed the API to look like this:

```js
redux: {
  rootReducers: {
   'SAY_HI': (state, action) => 'HI'
    'RESET': () => undefined,
  }
}
```

That way it can handle multiple rootReducers, and looks similar to regular reducers.

blairbodnar:
Very cool stuff @ShMcK ! I like it.

How about rather than returning `undefined`, the user has to explicitly set the state to the initial state. Perhaps changing the API to be like this:

```jsx
redux: {
  rootReducers: {
    'SAY_HI': (state, action, initialState) => 'HI',
    'RESET': (state, action, initialState) => ({ ...initialState }),
  }
}
```

What do you think? I could go either way, but I think I tend to lean this way because you can understand exactly what is going on when reading the line: 

`'RESET': (state, action, initialState) => ({ ...initialState }),`. 

When you read `'RESET': () => undefined` (without reading the docs), it might lead to confusion. Without reading the docs, one might interpret it as "Look like that sets the state to `undefined`"

Thoughts?

ShMcK:
A couple things:

- I don't think this API would be used that often. Its kind of in the "danger zone unless you're resetting the state" place. 
- I originally tried to piece out the initialState from the models, and then find a way to return it. It was kind of tricky, and involved passing models into the redux init. Then I thought: "what if someone adds dynamic models, model(something)", then their state wouldn't be added.

blairbodnar:
Good points...

For the second point, I was playing around this afternoon and got a version of it working with initialState as the third param. I will make a PR tomorrow if ya like? Or do you prefer we not go with that approach and keep it as the undefined will reset the state?

Also found a bug with this current PR and dynamic models:

1. init with a model
2. dynamically add another model `model({ ... })`
3. dispatch something that changes the state
4. dispatch a 'reset' (using the root reducer)
5. bug: the state isn't reset.

Hopefully that makes sense - sorry I'm in a rush and running out the door right now, will hopefully have some time tomorrow!

ShMcK:
I'd like to see the version with initialState as the third param. I can only see people using initialState in one circumstance though, if they want to reset state. So adding it as a param wouldn't be too helpful.

For the bug, add a test and and we can work it out.

Awesome ideas! APIs always end up better with some discussion.

blairbodnar:
Fixed the bug! :) 
https://github.com/rematch/rematch/pull/181/commits/af0356d30e2de45d15c208c282313c3d3be62bf7

PR #182, opened 2017-12-10
root reducers receive initial state as third argument

blairbodnar:
No need to accept this PR - We can kill it. I just thought I'd share an idea for a way to pass initialState as a third arg to the root reducer as discussed in #181


ShMcK:
That's pretty elegant! 

One possible issue: if a user specifies an initialState, it will be over written by the default state.

blairbodnar:
Ah nice catch!

PR #183, opened 2017-12-16
WIP - persist with RESET

ShMcK:
Currently not yet working. Consider this an outline.

Should allow RESET & PURGE of data with redux-persist.

@blairbodnar - maybe you can figure out why this triggers a crash.


ShMcK:
close in favor of #192.

PR #190, opened 2017-12-30
add basic demo to README

ShMcK:
A working demo on WebpackBin: https://www.webpackbin.com/bins/-L1cbpl4_RgEeGN1YkPS.

Still not sure if this is the best place to host a demo. I had issues previously with CodePen, but now I realize that was due to a bundling issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,8 @@ Dispatch can be called directly, or with the `dispatch.model.action(payload)` sh
 
 **React** | Vue | AngularJS | Angular 2
 
+- [Count Demo](https://www.webpackbin.com/bins/-L1cbpl4_RgEeGN1YkPS)
+
 ```jsx
 import React from 'react'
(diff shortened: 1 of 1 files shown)


blairbodnar:
Hmm, yeah WebpackBin seems to be acting a little flakey.

- I tried accessing the demo on Safari on OSX - It just sat at the WebpackBin loading screen.
- I tried again with Safari on OSX. It loaded this time and worked, but said "Script Error" in the bottom right.
- Tried it in Chrome, got the same WebpackBin loading screen hang.
- Tried it again in chrome after refreshing and it worked.

lol, so maybe not the best choice? Maybe we can get a CodePen demo working once the bundling problem is sorted out?

Thanks for taking a stab at this! 😎

ShMcK:
I agree, CodePen's a much easier way to go. Worth another go.

PR #206, opened 2018-02-09
Update listeners row to say hooks

ROB-64:
I am still learning this library (and really like it so far), but i think its hooks on the model right?


ROB-64:
Nevermind, but what are model.hooks used for in comparison to subscriptions?

PR #212, opened 2018-02-12
feat(meta): pass meta as third parameter to reducers

pierissimo:
## Motivation
rematch reducers signature is (state, payload) => state, 
whilst regular redux reducers signature is (state, action) => state

That make impossible for rematch reducers to access the 'meta' property of an action, a functionality that may be required in advanced scenarios (example: dynamic state paths)

## Possible solution
A backward compatible solution could be, to pass a third parameter 'ReducerContext' to rematch reducers.
Initially, the ReducerContext include just whole action object, but in the future maybe it could host additional informations.
What do you think?

Thank you

### DIFF
--- a/src/plugins/effects.ts
+++ b/src/plugins/effects.ts
@@ -1,5 +1,5 @@
 import { Dispatch, MiddlewareAPI, Store } from 'redux'
-import { Action, Exposed, Model, Plugin, PluginCreator } from '../../typings/rematch'
+import { Action, Exposed, Model, Plugin, PluginCreator, ReducerContext } from '../../typings/rematch'
 
 const effectsPlugin: PluginCreator = {
   expose: {
@@ -29,9 +29,11 @@ const effectsPlugin: PluginCreator = {
       })
     },
     middleware: <S>(store: MiddlewareAPI<S>) => (next: Dispatch<S>) => async (action: Action) => {
-          // async/await acts as promise middleware
+        const reducerContext: ReducerContext = { action }
+
+        // async/await acts as promise middleware
         const result = (action.type in effects)
-          ? await effects[action.type](action.payload, store.getState())
+          ? await effects[action.type](action.payload, store.getState(), reducerContext)
           : await next(action)
         return result
     },
--- a/src/plugins/effects.ts
+++ b/src/plugins/effects.ts
@@ -1,5 +1,5 @@
 import { Dispatch, MiddlewareAPI, Store } from 'redux'
-import { Action, Exposed, Model, Plugin, PluginCreator, ReducerContext } from '../../typings/rematch'
+import { Action, Exposed, Model, Plugin, PluginCreator } from '../../typings/rematch'
 
 const effectsPlugin: PluginCreator = {
   expose: {
@@ -29,11 +29,9 @@ const effectsPlugin: PluginCreator = {
       })
     },
     middleware: <S>(store: MiddlewareAPI<S>) => (next: Dispatch<S>) => async (action: Action) => {
-        const reducerContext: ReducerContext = { action }
-
         // async/await acts as promise middleware
         const result = (action.type in effects)
-          ? await effects[action.type](action.payload, store.getState(), reducerContext)
+          ? await effects[action.type](action.payload, store.getState(), action.meta)
           : await next(action)
         return result
     },
--- a/src/redux/reducers.ts
+++ b/src/redux/reducers.ts
@@ -1,6 +1,6 @@
 /* eslint no-underscore-dangle: 0 */
 import { combineReducers, Reducer, ReducersMapObject} from 'redux'
-import { Action, ConfigRedux, Model, Reducers, RootReducers } from '../../typings/rematch'
+import { Action, ConfigRedux, EnhancedReducers, Model, ReducerContext, Reducers, RootReducers } from '../../typings/rematch'
 
 let combine = combineReducers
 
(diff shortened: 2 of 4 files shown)


ShMcK:
We played with this idea earlier, but rather an actions "meta" object as a third parameter.

For example:

```js
dispatch.someModel.someAction(payload, meta) === dispatch({ type: 'someModel/someAction', payload, meta })
```

I agree, something like this is necessary.

pierissimo:
The idea behind a 'context' object is that is more scalable.
It would be easy and backward compatible to add feature in the future (example: rootState)

pierissimo:
Sure.
So, let's say that the next signature of the rematch reducer is:
```javascript
(state, payload, action) => state
```
In the future, if you want to add an additional parameter (to add functionality ),
you have to do:
```javascript
(state, payload, action, additionalVariable1) => state
/// and again
(state, payload, action, additionalVariable1, additionalVariable2) => state
```
and so on.


Instead, with the 'ReducerContext' approach, the signature could just be
```javascript
(state, payload, { action, additionalVariable1, additionalVariable1 }) => state
```

The 'rootState' example maybe it's not great, because it's an anti-pattern to access the root level state, from a  nested reducer.
Just an Idea, I'm more than happy with the 'action as third parameter' approach.

ShMcK:
I'm also happy with the "action" as the third parameter for effects & reducers.

I agree, rootState may be a bit dangerous, it's more modular if reducers know as little as possible about each other.

@blairbodnar what are your thoughts?

blairbodnar:
Hey guys, I'd be fine with having the 3rd parameter approach in this PR if you want to go ahead with it! 👍🏻

But it gets me thinking...  Maybe we should simplify even further and just have it as the 2nd parameter?

```js
// model
const count = {
  state: 0,
  reducers: {
    incrementBy: (state, action) 
    // 'action' is a normal redux 'action' object. So in this case:
    // { type: 'count/incrementBy', payload: 5, meta: 'stuff' }
    return state + action.payload
  }
}

// elsewhere
// rematch saves you having to specify the 'type'
dispatch.count.incrementBy({ payload: 5, meta: 'stuff' })
```

That would be a breaking change though... maybe worth thinking about for 1.0.0?

blairbodnar:
Hmm with my last comment... Maybe it make's rematch's `dispatch` no longer useful?

```js
dispatch.count.incrementBy({ payload: 5, meta: 'stuff' })
```
... Might as well just do it this way:
```js
dispatch({ type: 'count/incrementBy', payload: 5, meta: 'stuff' })
```

pierissimo:
Moving to a (state, action) => state signature means allowing regular 'redux' reducers to be compatible with models reducers.
It could be a good thing.

> Maybe it makes rematch's dispatch no longer useful?

I think that it remains useful when paired to the connect HOF provided by 'react-redux'.

```javascript
connect(
  {...},
  dispatch => ({ addOne: dispatch.count.addOne })
)
```
I was also thinking that could be useful to export from '@rematch/core' a 'effects' objects, that holds all the effects.
So that the code from above could be rewritten to:
```javascript
import { effects } from '@rematch/core'

connect(
  {...},
  { addOne: effects.count.addOne}
)
```

But that's another discussion.

blairbodnar:
Cool, yeah it could be a good thing! What do you think @ShMcK ?

And yeah, doesn't hurt to keep `dispatch.modelName.actionName/EffectName`. Like you said, it does make working `connect` easier.

Yes, I agree it would be useful to have easy access to the effects. Probably would also want to expose the 'normal' actions that the reducers generate  too.

```js
const count = {
  state: 0,
  reducers: {
    incrementBy: ...
  },
  effects: {
    incrementByAsync: ...
  }
 }
```

```js
import { actions, effects } from '@rematch/core'

connect(
  {...},
  {
    incrementBy: actions.count.incrementBy,
    incrementByAsync: effects.count.incrementByAsync
  }
)

```

@pierissimo Go ahead and make an issue for this and we can discuss more there 👍🏻

Thanks guys!

ShMcK:
As for `(state, action) => state` vs. `(state, payload, action) => state`, I'm not sure where I sit.

In my own code, I only use payload, and I found it annoying to always have to destructure out `{ payload }` on every call. I think this probably serves nearly all normal use cases.

On the other hand, having "payload", then action as a third param seems quite redundant. 

I think part of what a good framework does is force users into best practices, without limiting them. I don't see a reason for having access to the entire action, if a "meta" property could suffice.

If you consider a threshold for a breaking change, it has to be significantly better. I think I could be convinced 2 params are better than one, but I'm not there yet. 

I'm leaning towards either "meta" or "action" as a third param.

blairbodnar:
Good points @ShMcK . I also only use payload in my code.

Agreed with your points as well about a good framework guiding users towards the best practices.

My thoughts after digesting it a bit:

1. Probably not worth a breaking change.
2. I'd go with action as the third param.

ShMcK:
@pierissimo are you happy with meta?

On a related note, I just ran into a situation at work where I would have liked to have access to rootState.

pierissimo:
I've implemented the changes, now it's passing the meta as third parameter as discussed!

pierissimo:
@ShMcK ah!
What was the situation?

ShMcK:
I have a model for alerts & another for alertFilters.

I'd like to have access to `state.alerts` within the alertFilters model.

ShMcK:
We originally had an importable function called "getState", which was removed in an early beta.

I think that would re-solve the discomfort. It's true, models shouldn't be encouraged to interact with other models, but if need be, importing getState would make a developer aware it is happening.

Adding state to a meta object would work just as well. 

I'm okay to merge this PR now, but I'll leave it to @blairbodnar in case he has any thoughts or objections.

@pierissimo sorry again for the delay. What to do with the third param has been under discussion for a while without ever reaching a clear conclusion.

pierissimo:
@ShMcK 
No problem at all, the right decision requires thinking :)
The 'getState' function could be handy for sure, even though I'm not sure how easy can be testing with it, as mocking would be necessary.

ShMcK:
@pierissimo I'll put a PR together with `getState` this weekend and tag you on it.

blairbodnar:
No further thoughts here, go for the merge 👍🏻

ShMcK:
Published as @rematch/core 0.4

PR #214, opened 2018-02-13
fix(loading): allows errors to propagate

pierissimo:
The loading plugin seems to stop the propagation of errors that happens in the effects.
This fix is meant to solve it.


ShMcK:
Released as @rematch/core v0.3.4

pierissimo:
Thank you.
@ShMcK Is the npm publishing process part of a continuous integration routine?

ShMcK:
Not at this time. I'm not sure how to properly version releases with a CI, though I'm sure it can be done.

Plugins are all published separately as well, but maintained within the same repo. I'm not sure how these could be published and versioned correctly with a CI, unless it checks for the package.json version.

This would make a good issue.

PR #222, opened 2018-02-22
Fix/loading props proxy

pierissimo:
fix: the loading plugin is now proxying all the parameters
The bug was preventing the propagation of the rootState and meta objects


ShMcK:
Good catch. Will release as a patch on the loading plugin.

PR #225, opened 2018-02-28
select should be from @rematch/select

kaddopur:
`select` should be from `@rematch/select`, not `@rematch/core`


ShMcK:
Good catch. Select used to be part of core, but we pulled it out as a plugin.

PR #236, opened 2018-03-04
Reducers that can listen to other model actions

### LINKED ISSUE #231: Is it possible to have models act on actions within other model
Based on the API I couldn't find if this is possible. Scenario could be as follow:

I have a generic Loading Model that every time I load data (`reducers` or `effects` getting called with naming convention) it should change its state.

I have a User Model and a Project Model. Both of them implement the effect `load` and both of them implement the reducer `loaded` and `loadingFailed`.

ShMcK:
Closes #231. A question turned into a proposal.

Listeners should be able to capture actions for other models directly within reducers.

The implementation can change, it is still up for discussion and debate, just getting the ball rolling.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -117,6 +117,16 @@ An object of functions that change the model's state. These functions take the m
 }
 ```
 
+Reducers may also listen to actions from other models by listing the 'model name' + 'action name' as the key.
+
+```js
+{
+  reducers: {
+    'otherModel/actionName': (state, payload) => state + payload,
+  }
+}
+```
+
 ### effects
 
 `effects: { [string]: (payload, rootState) }`
--- a/test/listener.test.js
+++ b/test/listener.test.js
@@ -1,18 +1,18 @@
 describe('listener', () => {
-  test('should trigger state changes on another model', () => {
+  test('should trigger state changes on another models reducers', () => {
     const { init, dispatch } = require('../src')
     
     const count1 = {
       state: 0,
       reducers: {
-        increment: state => state + 1,
+        increment: (state, payload) => state + payload,
       }
     }
 
     const count2 = {
       state: 0,
       reducers: {
-        'count1/increment': state => state + 1
+        'count1/increment': (state, payload) => state + payload
       }
     }
 
@@ -20,7 +20,7 @@ describe('listener', () => {
       models: { count1, count2 }
     })
 
-    dispatch.count1.increment()
+    dispatch.count1.increment(1)
 
     expect(store.getState()).toEqual({ count1: 1, count2: 1 })
   })
--- a/test/listener.test.js
+++ b/test/listener.test.js
@@ -1,0 +1,27 @@
+describe('listener', () => {
+  test('should trigger state changes on another model', () => {
+    const { init, dispatch } = require('../src')
+    
+    const count1 = {
+      state: 0,
+      reducers: {
+        increment: state => state + 1,
+      }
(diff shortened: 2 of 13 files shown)


ShMcK:
Now a working implementation. Feel free to give it a try. See the test for an example.

ShMcK:
This will also lead to a solid update in the loading plugin.

Anything you might consider changing? Otherwise, will merge tomorrow.

fkrauthan:
Looks very good. The only other thing I was thinking about (but it might be overkill) to allow wildecards. E.g. `'*': () => ` listens to ALL events and reducers and `'test/*': () => ` listens to all events and reducers within the `test` model. That way you can probably turn the loader plugin into a 5 line project.

ShMcK:
Hmm... Can you provide a use case for wild cards?

The logic is already written within the subscriptions plugin, but I haven't found a use for it in my own code. It's also more costly to pattern match on every action dispatched.

fkrauthan:
I think the load plugin is a great use-case for it. Another use case could be a generic error tracer for example listening for `*/*error*` to remove some performance issues you could preprocess the model and only use pattern match when it contains `*`. But to be honest I don't have a full real usecase in mind it just popped into my head when looking over it.

ShMcK:
On that train of thought, it might be a good idea to have a more formalized way of handling errors. 
Worth putting some thought/research into.

fkrauthan:
Oh I just noticed that syntax might also be useful for `effects`. That I can have my own effect execute when another effect in a different model is called.

ShMcK:
TODO:
- [ ] handle dispatches from other model's effects

docs/api.md
```diff
@@ -117,6 +117,16 @@ An object of functions that change the model's state. These functions take the m
 }
 ```
 
+Reducers may also listen to actions from other models by listing the 'model name' + 'action name' as the key.
+
+```js
+{
+  reducers: {
+    'otherModel/actionName': (state, payload) => state + payload,
```
crazy4groovy: Technically, isn't this the "action type"?

ShMcK: Yes. In this case action types are all namespaced by the model name.

And you're right, it's not very clear. How should we phrase this to make it clearer?

crazy4groovy: This is a cool concept!

test/listener.test.js
```diff
@@ -0,0 +1,27 @@
...
+      reducers: {
+        'count1/increment': state => state + 1
+      }
+    }
+
+    const store = init({
+      models: { count1, count2 }
+    })
+
+    dispatch.count1.increment()
```
crazy4groovy: You may wish to pass in a payload here, and test that it comes through
`'count1/increment': (state, payload) => state + payload`

ShMcK:
I think will publish this as is. 

Multiple "effect" listeners will take a little longer to ensure they're 100%.

Multiple effects can be chained together, no issue there. but must handle several situations:

1. the root effect doesn't exist
2. the root effect is loaded after a listener effect is loaded
3. the listener effect is loaded after a root effect

_(Consider the effect on the model to be the "root effect")_

The order of effects fired is likely important. I'd like to see a solution that ensures the root effect fires first.

mgutz:
Must say this feature was the only thing holding me back. As soon as I saw this framework I thought, "cool but..." This framework has made redux fun instead of being tedious. Great job. Using it on a non-trivial project and so far so good.

PR #245, opened 2018-03-06
loading plugin counts the number of outstanding effects per effect, module, and global

crazy4groovy:
I was having problems with the loading module resolving a pending effect too early, because out of a batch of dispatched effects, the first completed one set the loading state to `false`.

This implements an integer counter instead of a boolean value.

Also cleaned up the plugin tests slightly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "gitbook-plugin-github": "^3.0.0",
     "gitbook-plugin-prism": "^2.3.0",
     "jest": "^22.3.0",
+    "redux-persist": "^5.9.1",
     "reselect": "^3.0.1",
     "rimraf": "^2.6.2",
     "rollup": "^0.56.2",
--- a/plugins/loading/README.md
+++ b/plugins/loading/README.md
@@ -62,11 +62,16 @@ init({
 
 ## Options
 
+### asNumber
+
+```js
+{ asNumber: true }
+
+The loading state values are a "counter", returns a number (eg. `store.getState().loading.global === 5`).
+
+Defaults to `false`, returns a boolean (eg. `store.getState().loading.global === true`)
+
 ### name
-
-The loading reducer defaults to the name of "loading".
-
-If you would like to change this, use the `name` option.
 
 ```js
 { name: 'load' }
@@ -74,9 +79,11 @@ If you would like to change this, use the `name` option.
 
 In which case, loading can be accessed from `state.load.global`.
 
+Defaults to the name of `loading` (eg. `state.loading.global`).
+
 ### whitelist
 
-A shortlist of actions. Named with "modelName" & "actionName".
+A shortlist of actions. Named with "modelName" / "actionName".
 
 ```js
 { whitelist: ['count/addOne'] })
--- a/plugins/loading/examples/react-loading-example/src/index.js
+++ b/plugins/loading/examples/react-loading-example/src/index.js
@@ -6,7 +6,7 @@ import createLoadingPlugin from '@rematch/loading'
 import App from './App'
 import example from './example'
 
-const loadingPlugin = createLoadingPlugin()
+const loadingPlugin = createLoadingPlugin({ asNumber: true })
 
 const store = init({
   plugins: [loadingPlugin],
--- a/plugins/loading/src/index.ts
+++ b/plugins/loading/src/index.ts
@@ -1,29 +1,28 @@
 import { Action, Model, PluginCreator } from '@rematch/core'
 
-const createLoadingAction = (show) => (state, { name, action }: any) => ({
-  ...state,
-  global: state.global + (show ? 1 : -1),
-  models: {
-    ...state.models,
-    [name]: state.models[name] + (show ? 1 : -1),
-  },
-  effects: {
-    ...state.effects,
-    [name]: {
-      ...state.effects[name],
-      [action]: state.effects[name][action] + (show ? 1 : -1),
-    },
-  },
-})
-
 interface LoadingConfig {
   name?: string,
   whitelist?: string[],
   blacklist?: string[],
 }
 
-const loadingPlugin = (config: LoadingConfig = {}): any => {
-  // validate config
+const createLoadingAction = val => (state, { name, action }: any) => ({
+  ...state,
+  global: state.global + val,
+  models: {
+    ...state.models,
+    [name]: state.models[name] + val,
+  },
+  effects: {
+    ...state.effects,
+    [name]: {
+      ...state.effects[name],
+      [action]: state.effects[name][action] + val,
+    },
+  },
+})
+
+const validateConfig = config => {
   if (config.name && typeof config.name !== 'string') {
     throw new Error('loading plugin config name must be a string')
   }
@@ -36,18 +35,26 @@ const loadingPlugin = (config: LoadingConfig = {}): any => {
   if (config.whitelist && config.blacklist) {
     throw new Error('loading plugin config cannot have both a whitelist & a blacklist')
   }
+}
+
+export default (config: LoadingConfig = {}): any => {
+  validateConfig(config)
 
   const loadingModelName = config.name || 'loading'
+
+  const hide = createLoadingAction(-1)
+  const show = createLoadingAction(1)
+
   const loading: Model = {
     name: loadingModelName,
     reducers: {
-      hide: createLoadingAction(false),
-      show: createLoadingAction(true),
+      hide,
+      show,
     },
     state: {
-      effects: {},
       global: 0,
       models: {},
+      effects: {},
     },
   }
 
@@ -107,5 +114,3 @@ const loadingPlugin = (config: LoadingConfig = {}): any => {
     }),
   }
 }
-
-export default loadingPlugin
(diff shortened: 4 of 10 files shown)


plugins/loading/test/loading.test.js
```diff
@@ -1,201 +1,231 @@
...
+  test('loading.global should be 2 for two dispatched effects', () => {
     const store = init({
       models: { count },
       plugins: [loadingPlugin()]
     })
+
+    dispatch.count.timeout()
     dispatch.count.timeout()
-    expect(store.getState().loading.global).toBe(true)
+    expect(store.getState().loading.global).toBe(2)
```
crazy4groovy: this is the gist of the new functionality

package.json
```diff
@@ -39,7 +39,8 @@
     "test": "jest"
   },
   "dependencies": {
-    "redux": "^3.7.2"
+    "redux": "^3.7.2",
+    "redux-persist": "^5.9.1"
```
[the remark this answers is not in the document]
ShMcK: I noticed that in your PR. Tests "should" pass with 'redux-persist' as a dev dependency.

plugins/loading/test/loading.test.js
```diff
@@ -1,201 +1,231 @@
...
-    const loadingPlugin = require('../src').default
+  test('should change the loading.effects to 2', () => {
+    const store = init({
+      models: { count },
+      plugins: [loadingPlugin()]
+    })
+
+    dispatch.count.timeout()
+    dispatch.count.timeout()
+    expect(store.getState().loading.effects.count.timeout).toBe(2)
```
crazy4groovy: this is the gist of the new functionality

plugins/loading/test/loading.test.js
```diff
@@ -1,201 +1,231 @@
...
+    const store = init({
+      models: { count },
+      plugins: [loadingPlugin()]
+    })
+
+    dispatch.count.timeout1()
+    dispatch.count.timeout2()
+    expect(store.getState().loading.effects.count.timeout1).toBe(1)
+    expect(store.getState().loading.effects.count.timeout2).toBe(1)
+    expect(store.getState().loading.global).toBe(2)
```
crazy4groovy: making sure global increments for every dispatched effect

plugins/loading/src/index.ts
```diff
@@ -55,45 +59,49 @@ const loadingPlugin = (config: LoadingConfig = {}): any => {
...
+          const origEffect = dispatch[name][action]
+
+          // create function with pre & post loading calls
+          const effectWrapper = async (...props) => {
+            try {
+              dispatch.loading.show({ name, action })
+              await origEffect(...props)
+              // waits for dispatch function to finish before calling "hide"
+            } finally {
+              dispatch.loading.hide({ name, action })
```
crazy4groovy: omit `catch`, but decrements counter even on an error

ShMcK:
Note that the loading plugin will likely require some changes once https://github.com/rematch/rematch/pull/236 is merged. Working on this at the moment.

At that point, loading can be handled with 2 actions, rather than the current 3.

1. show loading & action
2. hide loading

ShMcK:
@crazy4groovy - interesting idea - an integer count is certainly more useful than a boolean. I would use this, for its debugging value alone.

The only issue I can see is that users with typechecking (TypeScript, Flow) may find it annoying. 

Do you think it's possible to make it an "opt in" feature?

crazy4groovy:
@ShMcK interesting point about typechecking. Honestly I'm not strong on the topic so I don't have any insight, sorry. But I didn't see any `boolean` typing info that I was converting into `number` so maybe you can help me understand your concern with an example?

Boolean -> number is a breaking change, so opt-in would help maintain backwards-compatibility. That is often a smart move, unless the added feature completely trumps the existing implementation. Can you see any case where the new functionality is not better than current? Alternatively a v1.x (semver) bump may be the right move.

Thoughts?

ShMcK:
> Can you see any case where the new functionality is not better than current?

Nope. You have a good idea.

> Help me understand your concern with an example

In JS, there is no issue with dynamically typing variables.

```js
var loading = 0
if (loading) {
  // do something
}
```

But when using TypeScript, for example, the compiler expects variables to be statically typed.

```js
var loading = 0
if (loading) {          // warning, expected a boolean
  // do something
}
```

Then you have to do some nasty coercion to make the compiler happy.

```js
var loading = 0
if (!!loading) {
  // do something
}
```

---

I believe these two solutions can co-exist very simply.

1) store loading as a number
2) if users "opt in" to a config for "valueType" = "boolean", then the number is coerced into a boolean, but defaults to "number"

What do you think?

crazy4groovy:
> Then you have to do some nasty coercion to make the compiler happy.

I see. Yes, in JS world I would just expect to check loading like this:

```js
var loading = 0
if (loading > 0) { // always boolean
  // do something
}
```

I'm not opposed to a config param here, but it's not obvious to me how to store state as a number, and retrieve it as a boolean. Except maybe a selector of some kind? A hint here would be interesting for me.

crazy4groovy:
^^ poke?

ShMcK:
Sorry, I'll need a bit more time. I'll try to get back to you later this evening.

ShMcK:
I think both solutions can co-exist. I came up with something like this:

```js
const createLoadingAction = (valueFn) =>
  (show) =>
    (state, { name, action }: any) => ({
      ...state,
      global: valueFn[show](state.global),
      models: {
        ...state.models,
        [name]: valueFn[show](state.models[name]),
      },
      effects: {
        ...state.effects,
        [name]: {
          ...state.effects[name],
          [action]: valueFn[show](state.effects[name][action]),
        },
      },
    })

 let valueFn
 let defaultValue
  if (config.valueType === 'number') {
    valueFn = {
      hide: () => 0,
      show: (v) => v + 1,
    }
   defaultValue = 0
  } else {
    valueFn = {
      hide: () => false,
      show: () => true,
    }
    defaultValue = false
  }

  const loadingAction = createLoadingAction(valueFn)

  /* add defaultValue where needed */
```

In this case, users can opt in to using numbers over booleans by adding a `config.valueType = 'number'`.

This way, TypeScript/Flow folks are happy, no breaking changes, and the README can emphasize that numbers are an excellent option. What do you think?

crazy4groovy:
Interesting idea! Feels pretty complicated but it's an interesting concept, I'll try to take a look this weekend.

Also, I think the numerical `hide` isn't quire right, I think it should just tick down the number instead of setting it to 0.

ShMcK:
Great, also have a look at #266. Can you think of a way to resolve this with the loading plugin?

plugins/loading/src/index.ts
```diff
@@ -33,20 +51,32 @@ const loadingPlugin = (config: LoadingConfig = {}): any => {
...
     throw new Error('loading plugin config cannot have both a whitelist & a blacklist')
   }
+}
+
+export default (config: LoadingConfig = {}): any => {
+  validateConfig(config)
 
   const loadingModelName = config.name || 'loading'
+
+  const converter =
```
crazy4groovy: I'm pretty happy with this solution, it keeps a "background" counter called `cntState` of dispatched effects, but converts it onto its public state as either a number or boolean.

plugins/loading/test/loading-asNumber.test.js
```diff
@@ -0,0 +1,327 @@
+const delay = ms => new Promise(r => setTimeout(r, ms))
```
crazy4groovy: There are a lot more tests now.. I just duplicated them and tested for boolean in one file, and number in this one.

plugins/loading/README.md
```diff
@@ -62,21 +62,28 @@ init({
 
 ## Options
 
-### name
+### asNumber
```
crazy4groovy: yay for docs! :+1:

plugins/loading/examples/react-loading-example/src/index.js
```diff
@@ -6,7 +6,7 @@ import createLoadingPlugin from '@rematch/loading'
 import App from './App'
 import example from './example'
 
-const loadingPlugin = createLoadingPlugin()
+const loadingPlugin = createLoadingPlugin({ asNumber: true })
```
crazy4groovy: The example uses numbers, I think it's way more practical personally

ShMcK:
Quite an elegant solution with the "converter". Thanks for putting in all the effort @crazy4groovy, great work, and sorry this took so long. I really appreciate your patience.

ShMcK:
Getting a strange error from the Rollup bundler.

```
TypeError: parse is not a function in lib/index.js
```

WIll have to look at this tomorrow morning before I'm able to publish as `@rematch/loading@0.4`.

crazy4groovy:
Drat sorry. Here's a 100% shot in the dark to try: could be the `export default` statement I consolidated, try pulling it back out to the bottom maybe (I've just heard someone else mention that on a different project.)

crazy4groovy:
Maybe this issue: https://github.com/rollup/rollup-plugin-commonjs/issues/299

ShMcK:
Something changed in Rollup. Before I could use the peer dependencies in the root "rematch" directory, but now I had to install them all as dev dependencies in the plugins.

ShMcK:
Thanks for your help. Published as @rematch/loading@0.4.0.

semoal:
@tianzhich

PR #249, opened 2018-03-06
add example with vuejs-redux

titouancreach:
Hi,

I am the creator of [vuejs-redux](https://github.com/titouancreach/vuejs-redux). Vue/Redux binding that use scoped slots (native and documented feature of Vue). I forked the and edited the codesandbox to use my plugin and I added it in the readme. I'm not sure of the best integration in it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Rematch is Redux best practices without the boilerplate. No more action types, a
 * Recipes
   * [Devtools](./docs/recipes/devtools.md)
   * [React](./docs/recipes/react.md)
+  * [Vue](./docs/recipes/vue.md)
 * Plugins
   * [Subscriptions](./plugins/subscriptions/README.md)
   * [Selectors](./plugins/select/README.md)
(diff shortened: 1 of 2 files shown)


ShMcK:
Sorry this took so long. After comparing the two Vue demos, I much prefer your example.

Rather than binding a single store to Vue, your library also uses a Provider and passes in the store. I believe is more flexible, and more in line with where state management will be headed in the near future.

Could you change the PR to completely replace the current Vue demo link in the README instead?

Also, if you're interested, we could always use a Vue example under [recipes](https://github.com/rematch/rematch/tree/master/docs/recipes).

titouancreach:
Thank you for your feed back ! Vuejs-redux use scoped slots (a.k.a render props (or render children)) in react because HOC are not idiomatic in Vue and leads to a [lot of issues](https://github.com/vuejs/vue/issues/6201).
I'm also interested in making a recipe ! 

I do this as soon as possible.

titouancreach:
If you have any issues/questions in your repo that are related to the vue integration. You can tag me in it and I will be happy to resolve them.

PR #240, opened 2018-03-04
v1.0.0-alpha

ShMcK:
progress towards #239.

# Goals

1. TypeScript support
2. Support multiple stores
3. Dispatch into multiple stores with global dispatch

# Breaking Changes

1. global getState will now get state from all stores

PR #254, opened 2018-03-07
WIP - Experimental types

ShMcK:
Ongoing from the work of @rtfpessoa.

This PR aims to refactor Rematch to use strict typings.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4425,6 +4425,12 @@
         "symbol-observable": "1.2.0"
       }
     },
+    "redux-persist": {
+      "version": "5.9.1",
+      "resolved": "https://registry.npmjs.org/redux-persist/-/redux-persist-5.9.1.tgz",
+      "integrity": "sha512-WIaLCd975R16W7nD1wqUPLcqeLztIRvsSSNZdblEjtihL1wz+NSOq9nYUSvsdoWJOtfm9RbDF4IVd0a3VOtZcw==",
+      "dev": true
+    },
     "regenerator-runtime": {
       "version": "0.11.1",
       "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.11.1.tgz",
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "gitbook-plugin-github": "^3.0.0",
     "gitbook-plugin-prism": "^2.3.0",
     "jest": "^22.4.2",
+    "redux-persist": "^5.9.1",
     "reselect": "^3.0.1",
     "rimraf": "^2.6.2",
     "rollup": "^0.56.4",
--- a/src/core.ts
+++ b/src/core.ts
@@ -11,12 +11,9 @@ export default class CoreFactory<S> {
   constructor(model: ModelFactory<S>, localStore: LocalStore<S>) {
     this.model = model
     this.localStore = localStore
-
-    this.preStore.bind(this)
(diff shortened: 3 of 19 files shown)


ShMcK:
Some initial observations:

OOP has brought a lot of clarity to how things should be organized, and also made the typings very clear. You've also provided a clear way to provide a complete decoupling of stores.

I'll try to continue along this path to ensure tests pass, and the API remains consistent.

I can see a few additional feature possibilities thanks to the decoupling. It should be possible to have:
- multiple stores
- a global dispatch that can fire into multiple stores (though store.dispatch is still recommended)
- a way for stores to communicate with each other, if deemed necessary

@rtfpessoa feel free to continue, and thanks again for your contributions.

ShMcK:
43 / 139 tests passing 👍

nWidart:
I see the tsconfig file wasn't changed.

This could be added to it, to let typescript generate the type files.

```
"declaration": true
```

nWidart:
👀 Can't wait for this to work so I can use in a typescript project :D 

I'm thinking a "quick fix" could be to not provide types, which makes it usable in typescript if I'm not mistaken.

nWidart:
Hm, I'm not sure about that, possibly.

Though if there are no types at all, it could work. There are a lot of packages that don't provide types whatsoever, and it's still possible to use them in a typescript project.

However, if you're close to making this pull request work and that it'll make this package work with typescript projects, might not be worth it.

ShMcK:
Agreed. I'll put my focus into this PR today before falling back to plan B - no types.

resir014:
@ShMcK 

> Good call. Removing types and setting "declaration: true" would allow it to work in TS today?

CMIIW, but setting `declaration: true` will automatically generate a `.d.ts` file based on the types in your code, alongside your source file. You can then set `types` in your package.json to the generated types, in place of the custom typings.

I personally prefer having finer control over types with a custom `.d.ts` file over an automatically generated one, but your call!

nWidart:
I checked out this branch inside my typescript project, and sadly I'm still getting the error on my model.

image

yurifrl:
I seem to have the same issue here, changing ` export let dispatch: RematchDispatch` to ` export let dispatch: any` solves the issue for the time being

yurifrl:
This also seam to work 
```
export type RematchDispatchLiteral = {
    [key: string]: {
        [key: string]: (action?: Action) => Promise<Dispatch<any>>
    }
}

export type RematchDispatchCall  = (any)

export type RematchDispatch = RematchDispatchCall | RematchDispatchLiteral
```

nWidart:
Did you manage to override these types on your project without touching the package? 
That would be really nice, and allow us to use it now 😄

yurifrl:
No, I'm editing the package, and considering using a fork for now, I had an approach where I ignore the compiler with an assertion like this: `(<any>dispatch).count.increment()` but this caused an other issue, that I isolated here: https://stackblitz.com/edit/rematch-angular-5-count-whit-type-error?file=app%2Fapp.component.ts

yurifrl:
this seems to work `dispatch["count"]["increment"]()` with no refactor

ShMcK:
Currently working off of the 1.0.0-alpha.0 branch. Finding a slower migration with passing tests to be favorable.

Ultimate goal is: 
- multiple stores
- TS support

ShMcK:
close in favor of #240

PR #273, opened 2018-03-10
100% Test Coverage

kamerontanseli:
Added test for deprecate.ts pushing the coverage to 100%

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,15 @@
         "js-tokens": "3.0.2"
       }
     },
+    "@sinonjs/formatio": {
+      "version": "2.0.0",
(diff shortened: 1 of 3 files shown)


ShMcK:
Thanks @kamerontanseli. Good to see someone who finds test coverage equally important :)

One suggestion, and I may be wrong, but I believe jest can handle the stub in this case without requiring sinon as an additional dependency .

[An example](https://facebook.github.io/jest/docs/en/mock-functions.html#mock-return-values)

kamerontanseli:
@ShMcK I have replaced `sinon` with [`jest.spyOn`](https://facebook.github.io/jest/docs/en/jest-object.html#jestspyonobject-methodname)

ShMcK:
Thanks @kamerontanseli! Appreciate the contribution 👍

PR #275, opened 2018-03-11
A simpler example

g3r4n:
As highlighted in steps 3 we can use dispatch to make the example simpler

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,14 +135,14 @@ Dispatch can be called directly, or with the `dispatch[model][action](payload)` 
 
 ## Examples
 
-- Count: [JS](https://codepen.io/Sh_McK/pen/BJMmXx?editors=1010) | [React](https://codesandbox.io/s/3kpyz2nnz6) | [Vue](https://codesandbox.io/s/n3373olqo0) | [Angular](https://stackblitz.com/edit/rematch-angular-5-count)
-- Todos: [React](https://codesandbox.io/s/92mk9n6vww)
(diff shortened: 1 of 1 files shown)


ShMcK:
Good point. 

There are two goals to the example:
1) show a simple demo
2) show people how `dispatch`, `reducers` and `effects` work.

If we remove params from the example, users might not understand how to pass "payload" around.

g3r4n:
I didn’t change the steps 3 where there is an example about payload.
Should I correct something or close the PR ?

blairbodnar:
Thanks @g3r4n , It's funny - I debated a lot on this when we were doing those examples! In fact, I think we previously showed it the way you do here!

I personally like to use the `mapDispatchToProps` approach because it keeps the component totally isolated from rematch - It relies only on it's own props.

However, I do know that people like to pass the global dispatch directly like the PR suggests too!

Hmm, I'm torn. One one hand, it feels best practice-y to me to have the component pure... but on the other hand there is less code which shows off rematch's global dispatch. 

That being said, do you still feel the PR is the way to go?

g3r4n:
I' closing the PR because steps 1, 2, 3 and the example are showing different ways to work with this API. This PR will not bring more value than the current state.

ShMcK:
Thanks @g3r4n. I appreciate the PR, and certainly agree that your example is much simpler. Hopefully we can include easier ways to reduce boilerplate in a tutorial series for Rematch in the near future :)

blairbodnar:
Ditto, thanks @g3r4n !

PR #281, opened 2018-03-13
Add action creators to redux-devtools

jackwilsdon:
This PR adds action creators from each model to the Redux DevTools like so;

Redux DevTools

The action creators are generated from the reducers on the model;

```JavaScript
init({
    models: {
        counter: {
            state: 0,
            reducers: {
                increment: (state, delta = 1) => state + delta,
                decrement: (state, delta = -1) => state - delta,
            }
        }
    }
});
```

This PR makes it so that we add an `actionCreators` property to `config.redux.devtoolOptions` (containing the action creators) if the `actionCreators` property isn't already defined.

Marked this as WIP as I'm looking for some feedback on the changes :smile:

### DIFF
--- a/src/utils/getActionCreators.ts
+++ b/src/utils/getActionCreators.ts
@@ -1,0 +1,25 @@
+import { Models } from '../../typings/rematch';
+import isListener from './isListener';
+
+export default (models: Models): { [key: string]: Function } =>
+    Object.keys(models).reduce((actionCreators, modelName) => {
+        const { reducers = {} } = models[modelName];
+
+        Object.keys(reducers)
+            .filter(reducerName => !isListener(reducerName))
+            .forEach(reducerName => {
+                const type = `${modelName}/${reducerName}`;
+
+                // We have to dynamically create the function like this,
+                // so that the argument name is not minified.
+                const createCreator = new Function('type', `
+                    return function(payload) {
+                        return { type, payload };
+                    }
+                `);
+
+                actionCreators[type] = createCreator(type);
+            });
+
+        return actionCreators;
+    }, {});
--- a/src/init.ts
+++ b/src/init.ts
@@ -6,6 +6,7 @@ import corePlugins from './plugins'
 import { initReducers } from './redux/reducers'
 import { initStore } from './redux/store'
 import buildPlugins from './utils/buildPlugins'
+import getActionCreators from './utils/getActionCreators';
 import getExposed from './utils/getExposed'
 import getModels from './utils/getModels'
 import isObject from './utils/isObject'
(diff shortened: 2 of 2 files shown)


ShMcK:
Wow, I had no idea this was a redux devtools feature! I can see using it.

I'll have to play around with it before providing feedback.

jackwilsdon:
I've realised one issue with this, which is that it doesn't support lazy-loaded models (I don't think we can modify the Redux DevTools options once they've been set?).

ShMcK:
This issue may be resolved. See [devtools with replaceReducers](https://github.com/gaearon/redux-devtools/issues/16).

jackwilsdon:
I don't think that issue you linked is related? We need to be able to inform the devtools of the new actions when models are lazy-loaded, but there doesn't appear to be an API for that (it looks like once options are set, you can't change them?).

ShMcK:
You're right. There would have to be a way to replace enhancers in a Redux store.

jackwilsdon:
[It looks like there's a library for this which isn't all that complex](https://www.npmjs.com/package/redux-replaceable-middleware).

ShMcK:
Interesting. And very simple internally, this could be implemented: https://github.com/yldio/redux-replaceable-middleware/blob/master/index.js.

src/utils/getActionCreators.ts
```diff
@@ -0,0 +1,29 @@
...
+        const { reducers = {} } = models[modelName];
+
+        Object.keys(reducers)
+            .filter(reducerName => !isListener(reducerName))
+            .forEach(reducerName => {
+                const reducer = reducers[reducerName];
+                const type = `${modelName}/${reducerName}`;
+
+                // We have to dynamically create the function like this,
+                // so that the argument name is not minified.
```
ShMcK: Interesting. I hadn't thought about the effect of minification here.

jackwilsdon: Yeah it's a bit of a pain. I wrote the initial version of this without the dynamic function creation and ended up with `model/action(e)` in the dropdown of the devtool :confused:

ShMcK: Ouch. Learn something every day.

ShMcK:
@jackwilsdon - any last reservations before merging?

ShMcK:
Receiving the following error when dispatching a non-object. For example, a count example with a payload of 1.

```
Unhandled Rejection (Error): Actions must be plain objects. Use custom middleware for async actions.
```

jackwilsdon:
Looks like that was caused by actually calling `dispatch` in the action creator, whereas Redux DevTools expects the creator to just return an action to dispatch!

ShMcK:
All seems to be working now. Thanks!

PR #291, opened 2018-03-15
Remove extra quote

jackwilsdon:


ShMcK:
Strange. Thanks for going through this trouble.

PR #293, opened 2018-03-16
WIP: Add support for effect listeners

### LINKED ISSUE #246: Multiple Effect Listeners
Similar to #236 but for effects.

Multiple effects can be chained together, no issue there. but must handle several situations:

- the root effect doesn't exist
- the root effect is loaded after a listener effect is loaded
- the listener effect is loaded after a root effect

_(Consider the effect on the model to be the "root effect")_

The order of effects fired is likely important. I'd like to see a solution that ensures the root effect fires first.

jackwilsdon:
Fixes #246.

This is a preliminary implementation of this feature, and I'm not too sure if my implementation is the correct way of doing things (hence the WIP tag).

This PR also allows effects to be dispatched from actions in other models.

```JavaScript
const store = init({
  models: {
    auth: {
      state: { loggedIn: false, username: null },
      reducers: {
        loggedIn: (state, username) => ({ ...state, loggedIn: true, username }),
        loggedOut: state => ({ ...state, loggedIn: false, username: null }),
      },
      effects: {
        logIn: function({ username, password }) {
          this.loggedIn(username);
        },
        logOut: function() {
          this.loggedOut();
        },
      },
    },
    profile: {
      state: null,
      reducers: {
        loaded: (state, profile) => profile,
        'auth/loggedOut': () => null,
      },
      effects: {
        'auth/loggedIn': function(username) {
          // make AJAX request
          this.loaded({ username: username, age: 20 });
        },
      },
    },
  },
});

// triggers both auth/logIn and profile/auth/loggedIn effects
// output: {
//   auth: { loggedIn: true, username: 'jackwilsdon' },
//   profile: { username: 'jackwilsdon', age: 20 },
// }
store.dispatch.auth
  .logIn({ username: 'jackwilsdon', age: 20 })
  .then(() => console.log(store.getState()));
```


ShMcK:
Is this still a WIP ?

jackwilsdon:
I need to take another look at this, as it seems to have broken some things (and needs some tests). I won't be looking at it any time soon though, so it might be worth closing.

ShMcK:
I made a change to effects yesterday, so that they now process as actions. This should make this PR much easier. See changes for 0.6.0.

Baldrs:
@ShMcK but will this be supported in v1.0?

PR #303, opened 2018-03-22
Add documentation for using selectors with navigation plugin.

taranda:
Adds a mini tutorial on how to add selectors to the model generated by the react-navigation plug-in.  This is useful for determining the current route name.


ShMcK:
Nice, I may use this selectors technique as well.

Issue #302, opened 2018-03-19
Immutable JS support


taranda:
I am new to Immutable JS and Rematch.   My goal is  to using Rematch and Immutable JS together and make the entire state tree an Immutable JS object.  Is this possible?

According an author at [Redux.org](https://redux.js.org), one of the best practices regarding Immutable JS and Redux is to [make your entire Redux state tree an Immutable.JS object](https://redux.js.org/recipes/using-immutable.js-with-redux#make-your-entire-redux-state-tree-an-immutable.js-object).  They recommend replacing Redux's`combineReducers` with one  provided by [redux-immutable](https://github.com/gajus/redux-immutable).

I know rematch supports replacing the `combineReducers` method with  a custom method via the [Redux API](https://github.com/rematch/rematch/blob/master/docs/reduxApi.md#combinereducers).  However, will this affect the  entire Rematch state, or only apply to the reducers/state in the `redux:` section of the `init` configuration?

I would like  to take advantage of the `models` pattern using Immutable JS objects.

ShMcK:
I have not used Rematch with Immutable.js, though I believe this is possible. 

Replacing `config.redux.combineReducers` will replace all state with an Immutable.js state.

If you are looking for a simpler approach to using immutable state with Rematch/Redux, I would suggest trying [Immer](https://github.com/mweststrate/immer).

taranda:
Thanks.  I'll try that.  If I get it to work, I'll submit a documentation PR.

Issue #270, opened 2018-03-10
[select] support `rootState` select


leesiongchan:
In my case I have to connect to other states to complete my selector, is there any reasons why do we limit it to `model-level` only?

ShMcK:
Also noted in #276. 

@blairbodnar what are your thoughts on this one?

jackwilsdon:
Maybe something like this would work?

```JavaScript
const model = {
  selectors: {
    getData: (localState, rootState) => ...
  }
};
```

blairbodnar:
Hey guys! 

I think that we wrote the select plugin with the idea that it might be useful if you have _big_ models...

Let's say we have 3 counters. Maybe you choose to house all the logic in one model called `counters`. The select plugin might be useful here:

#### A bigger model:
##### models.js
```js
const counters = {
  state: {
    a: 1,
    b: 2,
    c: 3,
  },
  select: {
    sum: state => state.a + state.b + state.c
  }
}
```

#### Smaller models:
However, if you wrote it with 3 _smaller_ models, the select plugin wouldn't be much use...
##### models.js
```js
export const a = {
  state: 1
}

export const b = {
  state: 2
}

export const c = {
  state: 3
}
```
... but you could just write some selectors outside the model to bring your state together:
##### selectors.js
```js
export const sum = state => state.a + state.b + state.c
```

That being said...

@leesiongchan I would just write your own selectors outside of the models like the 2nd example above.
 (and not use the plugin). 😊

@jackwilsdon That's a good suggestion... but hmm... I'm thinking it might be best to keep it isolated and just write selectors outside of the model.

What do you guys think?

will-stone:
I feel that the select plugin could function without the need to be passed any parameters, i.e it’s always passed local state

jackwilsdon:
That makes sense, @blairbodnar. Maybe the `select` plugin can also offer a separate set of "global" selectors in that case?

```JavaScript
const store = init({
  models: {
    first: { state: 5 },
    second: { state: 10 }
  },
  plugins: [ select ],
  select: {
    total: ({ first, second }) => first + second
  }
});

store.select.total(store.getState()); // 15
```

will-stone:
@jackwilsdon I think getState is being deprecated.

leesiongchan:
Keeping it isolated sounds good to me.

jackwilsdon:
@will-stone Looking at the code, that only seems to be the root `getState` method called like so;

```

will-stone:
[...]

PR #310, opened 2018-03-26
Add support for Immutable JS in selector plugin.

taranda:
This PR allows the `select` plugin to be compatible with an Immutable JS store. As a side effect, it also allows the user to pass the root state to the selectors if desired.

This PR adds a config option to the `select` plugin allowing the user to specify how to slice the root state before being passed to the selectors  The default preserves current functionality so this change is complete backward compatible.  However, by opening up this config, a user can configure this plugin to support Immutable JS or another type of store.

This PR partially addresses Issue #270 and Issue #302.


taranda:
Adding my email/github is fine.  Not necessary, but I appreciate it.  I've added my email.  Thanks for the quick review. Cheers.

blairbodnar:
Released as @rematch/select 0.3.0

PR #313, opened 2018-03-28
Enhance loading plugin for Immutable.

taranda:
These enhancements make the loading plugin compatible with Immutable JS.  Two functions are exposed in the config which allow the user to customize how the state is initialized and updated. This allows the user to use an Immutable JS or other store that is not a JavaScript object.

This enhancement further adds a `model` config which allows the user to add additional customization to the `loading` model.  Perhaps the most useful use-case would be adding `selectors` to the model.

### DIFF
--- a/plugins/loading/README.md
+++ b/plugins/loading/README.md
@@ -98,3 +98,101 @@ A shortlist of actions to exclude from loading indicators.
 ```js
 { blacklist: ['count/addOne'] })
 ```
+
+### model
+
+`{ model: {...} }`
+
+A [model](https://github.com/rematch/rematch/blob/master/docs/api.md#model) configuration object which is included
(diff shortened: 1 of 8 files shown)


ShMcK:
@taranda I'm impressed you got this to work, while still leaving the loading plugin fully configurable. Great work.

ShMcK:
Published as @rematch/loading@0.5.0.

taranda:
Thanks...I still think there is room for improvement.  I am not sure how well Immutable JS will handle lazy loading models with `model()`, but it does work if the models are part of the initial config.  I've been thinking about removing the `mergeInitialState` config altogether and finding another way to initialize the state...perhaps taking advantage of the `onStoreCreated` API.

Anyway, thanks for creating such useful and clean framework for state management.  I am glad I can contribute.

PR #314, opened 2018-03-29
Added immer plugin

arcthur:


plugins/immer/README.md
```diff
@@ -0,0 +1,45 @@
...
+const immer = immerPlugin()
+
+init({
+  plugins: [immer]
+})
+```
+
+### Usage
+
+Reducer could be use mutable method to achieve immutable state, but this plugin may be some different that reducers must return value because literal don't support immer.
```
ShMcK: Does this mean that a state cannot have a primitive value?

plugins/immer/src/index.ts
```diff
@@ -0,0 +1,33 @@
+import { PluginCreator, RootReducers, Models } from '@rematch/core'
+import { combineReducers } from 'redux'
+import produce from 'immer'
+
+function combineReducersWithImmer(reducers: RootReducers) {
```
ShMcK: change type to `ReducersMapObject`

plugins/immer/README.md
arcthur: No, state can use primitive value, but must return changed value. because immer use observable to detect object or array change state. That means reducers don't return any change value, immer will be detected the change.

In there document, I just want to unify the reducer method.

ShMcK: So Immer only supports change detection on plain objects and arrays, but primitive values like strings or numbers will always return a change?

arcthur: Yes, that is observable limited, only to detect object or array.

ShMcK: Definitely worth mentioning then. Could we rephrase it in the README to say:

```
In Immer, reducers perform mutations to achieve the next immutable state. Keep in mind, Immer only supports change detection on plain objects and arrays, so primitive values like strings or numbers will always return a change.
``

File: plugins/immer/src/index.ts (hunk 2 above)
Comment (arcthur) [reply plugins/immer/src/index.ts]: ReducersMapObject don't export. How do I use?

File: plugins/immer/README.md (hunk 1 above)
Comment (arcthur) [reply plugins/immer/README.md]: good, I will modified the document.

File: plugins/immer/src/index.ts (hunk 2 above)
Comment (ShMcK) [reply plugins/immer/src/index.ts]: Try:

```js
import { ReducersMapObject } from 'redux'
```

arcthur: OK, that is worked. I committed.

ShMcK:
Thanks @arcthur, great work. Will publish as `@rematch/immer`.

PR #316, opened 2018-03-29
Add stateSlice method to react-navigation plugin config.

taranda:
This PR makes the react-navigation plugin compatible with Immutable JS.   The `sliceState` config  allows the user to configure how the navigation state can be found.  It also includes an example of how to use this plugin with Immutable JS.  There are no tests for this plugin but I did test the Immutable JS functionality on my own React Native app.

I would recommend testing for regression before merging.

It may just be my dev environment, but I could not get the package to compile without adding `"@types/react": "^16.1.0"`  to my `package.json` and removing `private` from the `Navigator` component's `render` method.   Those changes are not part of making this work for Immutable JS, but I could not compile without them.   Feel free to redact them if you see fit.


ShMcK:
Definitely should have made "render" `public`.

Thanks, another elegant solution.

ShMcK:
published as @rematch/react-navigation@1.1

ShMcK:
@taranda - any chance you could put together some notes on how to use Immutable.js with Rematch as a "recipe"?

PR #323, opened 2018-04-04
Add logic to create simple default reducers if none are provided

granmoe:
I've noticed that many models end up having a bunch of boilerplate, simple reducers of the form:

```javascript
setFoo: (state, foo) => ({
  ...state,
  foo
})
```

So I thought, "why not eliminate this boilerplate as well?" by using a simple convention. This PR makes it so that all keys in a model's state get a default reducer. The naming convention is to prepend "set" and uppercase the first letter of the key name. So, state.isLoggedIn would get a reducer named setIsLoggedIn. If the user provides a custom reducer, then this overrides the default.

Sorry about the formatting issues. I don't know why prettier is freaking out, but we can definitely get it straightened out if you guys decide you want to merge this PR. I figure I may as well get it out there early and get feedback since the basic concept is fairly simple.

### DIFF
--- a/src/redux/reducers.ts
+++ b/src/redux/reducers.ts
@@ -1,6 +1,13 @@
 /* eslint no-underscore-dangle: 0 */
-import { combineReducers, Reducer, ReducersMapObject} from 'redux'
-import { Action, ConfigRedux, EnhancedReducers, Model, Reducers, RootReducers } from '../../typings/rematch'
+import { combineReducers, Reducer, ReducersMapObject } from 'redux'
+import {
+	Action,
+	ConfigRedux,
+	EnhancedReducers,
+	Model,
+	Reducers,
+	RootReducers,
+} from '../../typings/rematch'
 import isListener from '../utils/isListener'
 
 let combine = combineReducers
@@ -9,58 +16,76 @@ let allReducers: Reducers = {}
 
 // create reducer for given dispatch type
 // pass in (state, payload)
-export const createReducer = (reducer: EnhancedReducers, initialState: any) =>
-  (state: any = initialState, action: Action) => {
-  // handle effects
-  if (typeof reducer[action.type] === 'function') {
-    return reducer[action.type](state, action.payload, action.meta)
-  }
-  return state
+export const createReducer = (reducer: EnhancedReducers, initialState: any) => (
+	state: any = initialState,
+	action: Action
+) => {
+	// handle effects
+	if (typeof reducer[action.type] === 'function') {
+		return reducer[action.type](state, action.payload, action.meta)
+	}
+	return state
 }
 
 // creates a reducer out of "reducers" keys and values
 export const createModelReducer = ({ name, reducers, state }: Model) => {
-  const modelReducers: Reducers = {}
-  Object.keys(reducers || {})
-    .forEach((reducer) => {
-      const action = isListener(reducer) ? reducer : `${name}/${reducer}`
-      modelReducers[action] = reducers[reducer]
-    })
-  return {
-    [name]: createReducer(modelReducers, state),
-  }
+	const upperCaseFirst = (str: String) => {
+		const [first, ...rest] = Array.from(str)
+		return `${first.toUpperCase()}${rest.join('')}`
+	}
+
+	const modelReducers: Reducers = {}
+	Object.keys(state || {}).forEach(key => {
+		const reducerName = `set${upperCaseFirst(key)}`
+		const defaultReducer = (state, payload) => ({ ...state, [key]: payload })
+		modelReducers[reducerName] = defaultReducer
+	})
+	Object.keys(reducers || {}).forEach(reducer => {
+		const action = isListener(reducer) ? reducer : `${name}/${reducer}`
+		modelReducers[action] = reducers[reducer]
+	})
+	return {
+		[name]: createReducer(modelReducers, state),
+	}
 }
 
 // uses combineReducers to merge new reducers into existing reducers
 export const mergeReducers = (nextReducers: Reducers = {}) => {
-  allReducers = { ...allReducers, ...nextReducers }
-  if (!Object.keys(allReducers).length) {
-    return (state: any) => state
-  }
-  return combine(allReducers)
+	allReducers = { ...allReducers, ...nextReducers }
+	if (!Object.keys(allReducers).length) {
+		return (state: any) => state
+	}
+	return combine(allReducers)
 }
 
 export const initReducers = (models: Model[], redux: ConfigRedux): void => {
-  // optionally overwrite combineReducers on init
-  combine = redux.combineReducers || combine
+	// optionally overwrite combineReducers on init
+	combine = redux.combineReducers || combine
 
-  // combine existing reducers, redux.reducers & model.reducers
-  mergeReducers(models.reduce((reducers, model) => ({
-    ...createModelReducer(model),
-    ...reducers,
-  }), redux.reducers))
+	// combine existing reducers, redux.reducers & model.reducers
+	mergeReducers(
+		models.reduce(
+			(reducers, model) => ({
+				...createModelReducer(model),
+				...reducers,
+			}),
+			redux.reducers
+		)
+	)
 }
 
-export const createRootReducer = (rootReducers: RootReducers = {}): Reducer<any> => {
-  const mergedReducers: Reducer<any> = mergeReducers()
-  if (Object.keys(rootReducers).length) {
-    return (state, action) => {
-      const rootReducerAction = rootReducers[action.type]
-      if (rootReducers[action.type]) {
-        return mergedReducers(rootReducerAction(state, action), action)
-      }
-      return mergedReducers(state, action)
-    }
-  }
-  return mergedReducers
+export const createRootReducer = (
+	rootReducers: RootReducers = {}
+): Reducer<any> => {
+	const mergedReducers: Reducer<any> = mergeReducers()
+	if (Object.keys(rootReducers).length) {
+		return (state, action) => {
+			const rootReducerAction = rootReducers[action.type]
+			if (rootReducers[action.type]) {
+				return mergedReducers(rootReducerAction(state, action), action)
+			}
+			return mergedReducers(state, action)
+		}
+	}
+	return mergedReducers
 }


granmoe:
Tests are passing.

Also, I'll be happy to add documentation for this.

granmoe:
@ShMcK Thanks!

I'd be happy to do it as a plugin provided a bit of guidance--I don't know how the plugin system works. Wouldn't we have to monkey-patch or replace the reducers.ts file that I modified here?

blairbodnar:
Nice @granmoe !

I thiiiiiink this could be done as a plugin. Hang tight, I'll take a look. 😀

blairbodnar:
Damn, I don't think it could be done with the current plugin API.

However, I think it could be done if our plugin `onModel` hook was changed so that a modified model could could be returned. @ShMcK, what do you think about allowing the `onModel` hook to be:
```js
onModel: model => newModel
```
It could allow a plugin to modify the model before rematch processes the model. 

If we did that, the following code should work as a plugin for @granmoe's idea:

😎

```js
const upperCaseFirst = str => {
  const [first, ...rest] = Array.from(str)
  return `${first.toUpperCase()}${rest.join('')}`
}

const makeDefaultReducers = model =>
  Object.keys(model.state).reduce((defaultReducers, stateKey) => {
    defaultReducers[`set${upperCaseFirst(stateKey)}`] = (state, payload) => {
      return {
        ...state,
        [stateKey]: payload
      }
    }
    return defaultReducers
  }, {})

export default () => ({
  init: () => ({
    onModel(model) {
      return {
        ...model,
        reducers: {
          ...makeDefaultReducers(model),
          ...model.reducers
        }
      }
    }
  })
})

```

ShMcK:
The issue seems to be that it's hard to determine the order that onModel hooks will run.

If you want to change the model itself, some kind of a `preOnModel` hook seems necessary. Naming things is hard.

granmoe:
@ShMcK @blairbodnar 

Is there anything further I should do on this PR?  Should I look into doing what you guys are talking about? It sounds like the plugin system itself would have to be changed.

blairbodnar:
@granmoe 

Had a chance to look into this - Looks like we don't need to make any changes to Rematch or the Plugin API at all!

### Solution 1: Modify the models before they are passed into `init`.
#### Check out a working example here: https://codesandbox.io/s/pkxj2lxyq
```js
import { init } from '@rematch/core'
import * as models from './models'

const makeModelsWithDefaultReducers = (models) => { ... }

const store = init({
  models: makeModelsWithDefaultReducers(models)
})
```

### Solution2: As a plugin.
Plugins may generate a config for rematch init 
```js
import { init } from '@rematch/core'
import * as models from './models'

const makeModelsWithDefaultReducers = (models) => { ... }

const defaultReducerPlugin = {
  config: {
    models: makeModelsWithDefaultReducers(models)
  }
}
const store = init({
  plugins: [defaultReducerPlugin]
})
```

So yeah, It's nice and easy to just have one of these solutions locally on your project. Does that work for you for now?

That being said, this is a really cool idea you've come up with here! I've been playing around with making it into a plugin with some more customization: Eg supporting different types of state, naming conventions, maybe whitelisting / blacklisting models. etc... Will make a PR once I've got something worth sharing! 😀

granmoe:
@blairbodnar Woohoo! 💯 🏆 🎉 That is awesome! Thank you for investigating this!

Can't wait to see that PR :-) I think I'll go with the first option in the meantime for the sake of simplicity...and also because I'm ignorant of how the whole plugin system works and integrates with rematch's architecture ;-)

Very cool to think of the possibilities you mentioned. Perhaps we could inspect initial state and have conventions based on that, too? For example, if a field has initial state of an array, you could create reducers that append an item or delete an item at a given index or find and delete according to a function passed into Array.findIndex. (Just improvising here.) Or if initial state is an object, create default reducers that update a given key within the object or delete it. All of these default reducers would follow a strict naming convention (e.g. an array in state with key name of "items" would get default reducers with names like "appendItem" "removeItemAtIndex" etc). 

Tons of possibilities here. So many that I think it would be good to encourage people to use and refine the default default reducer conventions ;-) while exposing something in the API to allow customizing default reducer logic if desired (could be as simple as what you've shown above).

granmoe:
Oh...I think I understand the plugin system now. Does the plugin's config just get merged into the default config via something like `const finalConfig = { ...config, ...pluginConfig }`? Does that mean plugins could override each other depending on order?

blairbodnar:
Thanks @granmoe! 😀

Great comments above - We're on the same page for sure! I had many of the same thoughts!

And yeah, the config that a plugin generates gets merged, so the order of the plugins can be important.

Ok I'll take a stab at a PR this afternoon. Stay tuned!

blairbodnar:
@scottclayton take a look when you have a sec. This is your jam! I know you've automated some reducer creation before 😎
Also @ShMcK let me know any thoughts when you have some time!

granmoe:
My initial thoughts: 

* The customizability is awesome! The plugin can provide the most commonly needed / obvious / least controversial (grasping for words here) reducers while providing an API that allows total control for less common / more contentious use cases :-) And I guess any PRs requesting that an implementation be added will just need to be considered on a case by case basis according to how much need there is in the community and the cost of adding it to the plugin.
* Do models need to be passed into `generateModelReducersPlugin`? I would have thought that the plugin would get passed the models by `init` so that the developer doesn't have to do this. You allow for a function to determine what models to apply the plugin to, so you're probably not passing them in order to do this.
* Great idea to allow passing a function to determine which models to apply the plugin to! Would each implementation optionally take this function as an argument?

My only other wish is that the specific default reducers I wanted (setKeyname) would be included in some implementation provided by the plugin. Could we have something like `implementation.setValue` / `implementation.modelsWithSet` (ambiguous because of the `set` data type...hmmm...naming things is always hard). Also...what happens if the implementations conflict with each other? How are they reconciled? Should it just be a matter of running them in sequence and overriding the previous result? I think that would work. And the more specific ones would by definition apply to fewer parts of state anyway, so it should work out fairly well.

granmoe:
Also...if the implementations are made to be somehow easily composable, it would be super easy to create a default setup like `const myImplementation = [withValues, withArrays, withObjects]` and then pass this into `generateModelReducersPlugin` 😀If a certain setup ends up being really common, we could add it to the plugin and either make it the default or export it. I'm not biased, but I would make having the simple `setKeyname`, `set` and `reset` the default ;-)

granmoe:
I like that you can pass in options to an implementation (like `setName` and `resetName` for `modelsWithObjectState`). With that level of customization, I think just a handful of these should cover 90% of the most common use cases.

blairbodnar:
Thanks @granmoe!

* Thanks! Yeah passing in options should hopefully handle 90% of use cases 😎
* Agreed on adding community implementations - I was thinking the same -  just do them on a case by case basis - have people put in a quick and easy PR for implementation.
* `myImplementation = [withValues, withArrays, withObjects]` - I was thinking the same! Reminds me of the [babel presets ](https://babeljs.io/docs/plugins/#presets), haha.
* Agreed that it would be nice to not have to pass the models manually into the plugin. Unfortunately the current API `0.6.0` and the `1.0.0-alpha.0` we just put out doesn't  allow allow for this. That being said, after playing with the alpha plugin API, I think there might be some improvements we can do that can allow for this (I’ll try to look into that more in the evenings this week) 👍🏻

scottclayton:
Hey @blairbodnar @granmoe 

I do love automating reducers—awesome to make it a plugin! I found that a single reducer `set` reduced my number of reducers by about two thirds. Actually, on an app I'm working on recently, I have written zero reducers other than the ones I bootstrap onto each model! I always make my models as objects, and historically have put my models into an array and forEach'ed over them before initializing the app. I do the following three reducers generally:

- **set**—creates a reducer that I call with the changes I'm making to a model's state's keys
    - e.g. `dispatch.profile.set({ username: data.username, lastUpdated: new Date() })` would take my existing model and only update the `username` and `lastUpdated` keys from it
- **toggle**—creates a reducer that I pass a string or array of strings that correspond with keys I want to toggle from true/false or vice versa.
    - I haven't come up with a more elegant way of calling it other than something like `dispatch.auth.toggle({ keys: 'loggedIn' })`
- **resetToInitialState**—returns the model to whatever its initial state was

I put a clause on each function to not overwrite reducers if they already have one of those names. I don't worry about whitelisting/blacklisting models because I only add the three reducers and avoid overwriting reducers. It might not be unreasonable to want to whitelist/blacklist certain models.

In some other rematch plugins, they get config objects. I think something like that could work well in this conext. This is just an idea, based off the one previously mentioned a few messages above:

```js
import { init } from '@rematch/core'
import * as models from './models'
import generateModelReducersPlugin from '@rematch/generate-model-reducers'

const generateModelReducersConfig = {
  reducers: {
    set: (state, { type, ...restOfPayload }) => ({ ...state, ...restOfPayload }),
    resetToInitial: (state) => state,
    toggle (state, payload) => ({ ...state, ...etc }),
    // ^ the function I generally use for this is long and probably inefficient
  },
  blackListModels: ['auth'],
  // ^ the models to exclude… maybe it's whitelisting and not blackListing… I get the two confused often
}

const store = init({
  plugins: [
    generateModelReducersPlugin(generateModelReducersConfig)
  ],
})
```

There could be some common reducers in the plugin, e.g.
```js
import generateModelReducersPlugin, { set, resetToInitial, toggle } from '@rematch/generate-model-reducers'
```

I wrote the `set` reducer in that way for quick illustration—one would probably want it to have more flexible type-handling for arrays, as not everyone does objects for the state of their models.

e.g. (?)
```js
    set: (state, payload) => {

      // I'm just guessing at the type-checking stuff
      if (Array.isArray(state)) return payload
      if (typeof state !== 'object') return payload

      const { type, ...restOfPayload } = payload

      return { ...state, ...restOfPayload }

    },
```


I don't know enough about immer patterns to comment on any conflicts with the immutable plugin.

I'm curious @granmoe … what are the advantages to having `setProperty` reducers for each key in each model rather than a generalized `set` reducer? I've found it nice to have one I can update multiple keys on simultaneously, and it's not any trickier to debug since I can look at the payload and see what I was doing, plus I save hundreds of extra reducers in a large app. However, I haven't looked into other people's use-cases for this sort of thing yet.

granmoe:
@scottclayton This is excellent info!

Let me respond to your question to me:

> I'm curious @granmoe … what are the advantages to having setProperty reducers for each key in each model rather than a generalized set reducer? I've found it nice to have one I can update multiple keys on simultaneously, and it's not any trickier to debug since I can look at the payload and see what I was doing, plus I save hundreds of extra reducers in a large app. However, I haven't looked into other people's use-cases for this sort of thing yet.

Come to think of it, this is just a pattern I saw when first learning redux a few years back, and I've just always stuck with it. Your way is simpler, but I immediately worry that it will make it hard to see what's happening in the redux devtools. I'll try to list out how I see the benefits of each approach:

# set

- less code inside of `connect` in components in order to wire up action creators since you just have one `set`
- default reducer generator code would be simpler
- don't need to write "batched" action creators to update multiple things at once
- maybe more of a predictable convention since you just have one `set` per model

# setProperty

- easier / faster to debug with redux devtools since you can verify that you see a series of certain action types as opposed to having to inspect each action's payload
- avoids giving too much power to the caller since `set` can do anything, as opposed to explicitly updating one thing at a time or creating "batched" update action creators if/when needed
- more explicit / specific actions (I guess this one is a little vague)
- wouldn't need to alias `set`s to something to avoid name collisions if you use multiple in one component (e.g. `connect(null, { profile: { set: setProfile }, auth: { set: setAuth } })(Component)`) (this isn't really that bad, but it's worth noting)

I'm very open on this. I just wanted to put my initial thoughts out there. I think your way is definitely more concise and convenient (and would simplify the internal rematch code to generate default reducers), but I'm trying to play devil's advocate for the other approach to make sure we don't overlook anything :-)

granmoe:
@scottclayton @blairbodnar I'm torn on this one. I think maybe this is a good case in which to support both options. What do you think? Either have one implementation that covers both options and let's you configure them to some extent, or have an implementation for each option.

In the meantime, I think I'm going to create branches for both options at work and see what our code looks like with either one :-)

granmoe:
Hey @scottclayton, I thought of one other situation where you wouldn't want to use `set`. If you have a few explicit `setProperty` reducers that have some special logic for some property in your state, then you could (mainly just if you're working in a large, team project) run into an issue where `set` is used to set that property in a different way than the explicit `setProperty` does, resulting in conflicting reducer logic and potential bugs. So I guess it's probably best not to use both ways at the same time for this reason!

I think this would also be a problem if you had any special logic to transform any parts of the payload before passing to `set` (say you *do* have this logic but just call it inline or from a utility function or something). This kind of supports my intuition that `set` has too much power :-) I think for this reason, I would only use it for small / solo projects.

That said, I would definitely still prefer to have `set` available and use it for the appropriate use cases. It is simpler.

blairbodnar:
Hey @granmoe, sorry for the delay on this one.

I had a look at the plugins API, and the current design (and 1.0.0 design) won't  allow the model to be modified before it makes it's way into rematch. :( 

I chatted a bit with @ShMcK and @scottclayton (we work together) about this and we all seemed kind of on the fence about changing the plugin API to allow models to be modified as part of the plugins API right now. 

In the meantime, we could just make the ideas discussed here into a recipe for auto-creating reusable reducers since it really doesn't require a change to rematch or having a plugin at all - It's just a technique a developer can use which involves modifying their models before passing them into rematch.

If that's cool we can close off this PR and make a recipe for it?

Thanks!

granmoe:
@blairbodnar Yeah, that all makes sense :-) We already started using that approach in our project, and it only takes a tiny amount of code. Now our rematch models are super streamlined.

Anyhow, thanks for entertaining this idea. The discussion has been awesome :-)

blairbodnar:
Awesome, glad to hear it's working out nicely! Thanks @granmoe! 😊

derrickbeining:
@blairbodnar, @granmoe , @scottclayton , @ShMcK , Hi everyone 👋 

Really enjoyed reading this discussion. I'm definitely in favor of the `setProperty` convention over plain `set`, but sounds like nothing actually ended up being implemented for users of `rematch` to provide some opt-in default reducers. 

A plugin may not be viable, but is there any interest in having `rematch` ship with its own `withDefaultReducers` utility that is named-exported along with the `init` function? Something that could be used along these lines:

```javascript
import { init, withDefaultReducers } from '@rematch/core'

const store = init({
  models: withDefaultReducers({
    ObjectModel: { state: { ...props } },
    ArrayModel: { state: [ ...elements ] },
    PrimitiveModel: { state: <Boolean|String|Number> }
  })
})
```
Or perhaps:
```javascript
const store = init({
  models: {
    ObjectModel: withDefaultReducers({ state: { ...props } }),
    ArrayModel: withDefaultReducers({ state: [ ...elements ] }),
    PrimitiveModel: withDefaultReducers({ state: <Boolean|String|Number> })
  }
})
```
Or maybe both options.

I think it'd be really great for `rematch` to provide a bunch of built-in reducer actions with conventional names generated based on models' initial state that users can easily opt in or out of. It would make working with Rematch/Redux about as streamlined as it can be.

I'm working on a fairly heavy-duty implementation of `withDefaultReducers` for my own purposes at the moment (provides a lot more than just `setProp`) and just wanted to see if there's any interest in having something like that internal to the `rematch` library, or if the consensus is to just let folks in user-land create their own `withDefaultReducers` implementations if they want it.

granmoe:
@derrickbeining I agree. I'd love to see this as an opt-in part of rematch. I agree it would make rematch all the more appealing.

I have a minimal version of makeDefaultReducers that I'm using in my project at work. It's very similar to what @blairbodnar had in one of those codesandbox demos up there.

granmoe:
The only thing I create in my version is setKeyName for every key in state, and a reset reducer that resets to initial state.

derrickbeining:
Published `rematch-default-reducers` on npm, if anyone's interested: https://www.npmjs.com/package/rematch-default-reducers

It also includes optional run-time type checking. So default reducers will throw an error if you try to set something on a model that is not of the same data type as what you initialized it with.

blairbodnar:
Cool! Thanks @derrickbeining

Issue #322, opened 2018-04-03
Unable to subscribe to actions from redux.reducers


miickel:
My init looks like this:

```javascript
let store = init({
  redux: {
    reducers: {
      router: router.reducer,
    },
    middlewares: [router.middleware],
    enhancers: [router.enhancer],
  },
  models,
})
```

Here `router` comes from [redux-little-router](https://github.com/FormidableLabs/redux-little-router).

My model looks like this:

```javascript
export const model = {
  state: {},
  reducers: {
    'ROUTER_LOCATION_CHANGED': (state, payload) => {
      // never trigger
    },
  },
}
```

I can see using Redux dev tools that `ROUTER_LOCATION_CHANGED` gets dispatched, yet my rematch reducer function never gets called.

I'm able to subscribe to rematch model actions using the `{model}/{reducer}`-convention, but can't find any way to subscribe to actions dispatched outside of rematch.

Is this supported?

ShMcK:
I think there are a few options here:

1) use "redux.reducers" instead of models.
2) there is a subscriptions plugin, but I think this will get deprecated in the near future.

---

## Proposal

I think a more elegant solution would be if you could map actions to other action names.

```js
export const model = {
  state: {},
  mapActions: {
    'ROUTER_LOCATION_CHANGED': 'routeChange'
  }
  reducers: {
    'routeChange': (state, payload) => {
      // never trigger
    },
  },
}
```

This could be made possible in core or through a plugin. What are your thoughts?

miickel:
I like your mapActions proposal, but perhaps mapping on the model level is a bit too granular? I mean, if I want to map ROUTER_LOCATION_CHANGED to routeChange, wouldn't that be true for all my models in that project?

Another proposal is to simply namespace the actions `@redux/ROUTER_LOCATION_CHANGED `

ShMcK:
That proposal sounds much simpler. @blairbodnar what are your thoughts?

blairbodnar:
I think giving a namespace to those actions is a great idea @miickel! I like how it follows the convention redux uses in [action types](https://github.com/reactjs/redux/blob/master/src/utils/actionTypes.js). 

```js
export const modelA = {
  reducers: {
    'set': // reducer for action type: 'modelA/set' (this model)
    'modelB/set': // reducer for action type: 'modelB/set' (another model)
    '@@redux/something': // reducer for action type: '@@redux/something' (not a model)
  }
}
```
I like it!

[...]

PR #324, opened 2018-04-04
feature/subscribe-model-reducers-to-non-model-actions

blairbodnar:
As discussed in #322

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -116,7 +116,7 @@ An object of functions that change the model's state. These functions take the m
 }
 ```
 
-Reducers may also listen to actions from other models by listing the 'model name' + 'action name' as the key.
+Reducers may listen to actions from other models by listing the 'model name' + 'action name' as the key.
 
 ```js
 {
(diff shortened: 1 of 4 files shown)


blairbodnar:
Wierd. All tests passed locally, not sure why `plugins/loading/test/loading-asBoolean.test.js` fails on travis. 🤔

blairbodnar:
Ugh, and didn't realize `plugins/persist/package-lock.json` was out of sync for me and made it's way into this PR.

blairbodnar:
... or don't bother with this, a nice middleware solution from @miickel solves it too. https://github.com/rematch/rematch/issues/322#issuecomment-378520145

ShMcK:
I like @miickel's solution as well. Do you think we should move into core?

blairbodnar:
Yeah, I think it'd be cool to have in core 👍🏻

miickel:
One drawback with my solution is that subscriptions in `effects` won't work. But perhaps that is by design?

blairbodnar:
Sorry for the delay on this one... Maybe we should just make @miickel's middleware solution https://github.com/rematch/rematch/issues/322#issuecomment-378520145 into a recipe?

blairbodnar:
Closing this off, we can make this into a recipe instead. https://github.com/rematch/rematch/issues/322#issuecomment-378520145 Thanks!

PR #335, opened 2018-04-10
Fix the documentation for DevTools options

sarunast:
The options should be passed down to `redux` object. Not on `init` object.


ShMcK:
Thanks for catching this.

Issue #334, opened 2018-04-09
DevTools config doesn't work 


sarunast:
I was trying to config the `devtoolOptions` by setting the name and other options. However, I don't see any changes. Using `"@rematch/core": "^0.6.0"`.

```js
const store = init({
  devtoolOptions: {
    name: 'MyApp',
  },
  models: {},
})
```

sarunast:
I figured it out the documentation was incorrect. I made a PR by fixing the documentation. https://github.com/rematch/rematch/pull/335

PR #336, opened 2018-04-10
v1 - Feature/add store name to devtools

ShMcK:
Automatically adds store name to devtools. Largely for use with multiple stores.

```js
const store = init({
  name: 'store-name',
})
```

Uses store.name as devtools.name


ShMcK:
Inspired by @sarunast 's recently closed issue #334.

Issue #282, opened 2018-03-13
Recipe for byId allIds pattern


titouancreach:
Hi,

I'm converting my entire codebase from redux to rematch :+1: 
For each entities, I has a state like:

```javascript
const state = {
  byId: {
    '3': {
      //myEntityHere
    }
  },
  allIds: ['3']
};
```
I used everytime `combineReducer` to separate my: ById reducer and my AllIds reducers.

To refactor this in rematch, why would be better:
  - Create a lot of models like: myEntityAllIds and myEntityById
  - put the the byId and allIds in the same model (but the reducer will look ugly)
  - other ?

Thanks :+1:  

I think this is related to: 
https://github.com/rematch/rematch/issues/230

jackwilsdon:
In your case I'd put them in the same model, as it's tightly related data.

```JavaScript
const model = {
  state: {
    byId: {},
    allIds: []
  },
  reducers: {
    load: (state, data) => ({
      byId: data, // Assuming your data is already keyed by ID.
      allIds: Object.keys(data)
    })
  }
};
```

If your data isn't already keyed by ID, it still shouldn't make the reducer too complex;

```JavaScript
const model = {
  state: {
    byId: {},
    allIds: []
  },
  reducers: {
    load: (state, data) => {
      // Assuming data is an array of objects with an "id" attribute.
      const byId = data.reduce((dataById, item) => ({ ...dataById, [item.id]: item }), {});

      return { byId, allIds: Object.keys(byId) };
    }
  }
};

**Comment by titouancreach** [comment]
Yes! Thanks for answering, I went that way. This is maybe a minor but it has changed my reducers from:

```javascript
 updateType(state, {entityId, type}) => ({
    ...state,
    [entityId]: {
      ...state[entityId],
      type
    }
  })
```

to
```javascript
    updateType(state, {entityId, type}) {
      return {
        ...state,
        byId: {
          ...state.byId,
          [entityId]: {
            ...state.byId[entityId],
            type
          }
        }
      };
    },
```
one level deeper. Even if the code is not that much complicated, it could lead to humain errors :(

I would imagine something like:


```javascript
reducers: {
  byId: {
    updateType(state, payload) {
      // state is localState.byId
    }
    add(state, payload) {
      return {
        ...state,
        [payload.id]: payload
      }
    }
  },
  allIds: {
    add(state, payload) {
      // state is my array
        return [...state, payload.id]; // add my id to the list 
    }
  }
}
```

[...]

PR #339, opened 2018-04-10
decoupling reducers recipe

titouancreach:
From: https://github.com/rematch/rematch/issues/282


ShMcK:
great work! I hope you don't mind if I act like an editor for a bit within this PR.

ShMcK:
What do you think about describing "Decoupling Reducers" as "Reusable Reducers"?

My understanding is "decoupling" refers more to separating two parts, so that neither part knows about each other. The pattern of turning reducers into reusable pure functions doesn't strike me as fitting this definition. What do you think?

docs/recipes/decouplingReducers.md
```diff
@@ -0,0 +1,141 @@
+# Decoupling reducers
+
+Rematch and more generally Redux, encourage to keep a flat state to solve several [issues](https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape).
```
ShMcK: Rematch, and more generally Redux, encourage you to...

docs/recipes/decouplingReducers.md
```diff
@@ -0,0 +1,141 @@
...
+Each model should be considered as a separate entity and should only be updated by its reducer, this is why Rematch doesn't allow nested reducers.
+
```
ShMcK: Nested reducers can be used in Rematch, but are not recommended. I think that is an important distinction.

Some users have set up functions, similar to combineReducers, that generate their nested reducers. I've also seen people use redux.reducers with combineReducers.

docs/recipes/decouplingReducers.md
```diff
@@ -0,0 +1,141 @@
...
+This being said, there is some case where nested reducer can be a good thing.
```
ShMcK: nested reducer**s**

docs/recipes/decouplingReducers.md
```diff
@@ -0,0 +1,141 @@
...
+In this recipe, we will see how to decoupling reducers that tends to grow too much.
+
```
ShMcK: In this recipe, we will see how to use reusable reducer functions.

docs/recipes/decouplingReducers.md
```diff
@@ -0,0 +1,141 @@
...
+Let's start with an example. A common pattern in Redux, when we want to store an ordered list, is "allIds, byId". Store  the entities as
+an array requires to iterate over the array to find our target but store the entities as an object doesn't let us to keep 
+an order. So let's use both:
```
ShMcK: "Storing the entities as an array requires an additional cost of iterating over the array to find our target. On the other hand, storing the entities as an object doesn't preserve the order of items"

docs/recipes/decouplingReducers.md
```diff
@@ -0,0 +1,141 @@
...
+      1: {
+        task: 'Learn functional programming',
+        isDone: false
+      }
+    },
+    allIds: [1, 0]
+  }
+}
+```
+`byId` key stores the entity while `allIds` keeps the track of the order of our entities.
```
ShMcK: "keeps track"

docs/recipes/decouplingReducers.md
```diff
@@ -0,0 +1,141 @@
...
+
+This driven us into an edge case:
```
ShMcK: This creates another issue:

docs/recipes/decouplingReducers.md
```diff
@@ -0,0 +1,141 @@
...
+   }
+  }
+}
+
+function toggleAllIds(state, payload) {
+  return state;
+}
+```
+
+And we finally mix all together by distibuting a part of the state to our update functions:
```
ShMcK: "finally mix everything all together"

docs/recipes/decouplingReducers.md
```diff
@@ -0,0 +1,141 @@
...
+      return {
+        byId: toggleById(state.byId, payload),
+        allIds: toogleAllIds(state.allIds, payload)
+      };
+    }
+  }
+};
+
+```
+Our model seems much better and more readable.
```
ShMcK: "Reusable reducer functions make our model simpler and more readable."

titouancreach:
> Consider these comments largely suggestions to help improve the readability. Overall, the tutorial is quite clear and easy to follow.

Don't worry, I know my writing in English is quite bad and I try to improve it! 

So I take note of what you say to become a better writer :)

titouancreach:
> What do you think about describing "Decoupling Reducers" as "Reusable Reducers" or "Reusable Reducer Functions"?

> My understanding is "decoupling" refers more to separating two parts, so that neither part knows about each other. The pattern of turning reducers into reusable pure functions doesn't strike me as fitting this definition. What do you think?

I think that our case here:
  - allIds doesn't know about byId because we pass state.allIids as an argument
  - byId doesn't know about allIds

docs/recipes/decouplingReducers.md
titouancreach: I would say, this is not possible out of the box: https://github.com/rematch/rematch/issues/282#issuecomment-372644100

titouancreach:
By the way, I didn't test the code I wrote, everything compile in my head, I would probably test it before we merge it

ShMcK:
Definitely give it a test run, if you don't mind. A link to a codesandbox.io demo always helps.

titouancreach:
https://codesandbox.io/s/lry6024mkl This is the example without decoupling (I've to change in the model the key "reducer" to "reducers"

titouancreach:
https://codesandbox.io/s/x2r7nryn24 the second example with decoupling reducers. The mistake I made

Spell "toogle" instead of "toggle"
declare: filterObjectByKey and call filterObjectById

titouancreach:
LGTM, open to feedbacks

ShMcK:
Is it alright if I change the file name to "resuableReducers.md" ?

titouancreach:
If we do that, we need to change the title too. For me, decoupling reducer is good because it's what we are doing

titouancreach:
Thank you :)

PR #357, opened 2018-05-07
Improve TypeScript integration / More autocomplete

### LINKED ISSUE #356: [PROPOSAL] Improve TypeScript integration
I'm playing with `rematch` and it's great, but the integration with TypeScript seems to be in early stages and harder than it should be. For example, how to make the autocomplete work when using the redux `@connect` `mapState` and `mapDispatch`?

The `select` object from the `@rematch/select` import gets magically filled with the the model names and selectors, plus it changes the function signature, both complicates things.

---

I have an idea that MIGHT help, bare with me.

Currently the model is declared like this:
```ts
export const app: Model = {
  state: {
    theme: 'dark',
  } as State,
  reducers: ({
    setTheme(state: State, theme: Theme) { },
  }),
  selectors: {
    getThemeObject(state: State) { },
  },
}
```

When using redux `@connect`, we do:
```tsx
const mapState = state => ({
  theme: select.app.getThemeObject(state), // no autocomplete
})

const mapDispatch = ({ app: { setTheme } }) => ({ // no autocomplete
  setDarkTheme: () => setTheme('dark'), // no autocomplete
})
```

---

What if we did something like this instead:

```ts
export const app = {
  state: {
    theme: 'dark',
  } as State,
  reducers: (state: State) => ({ // <-
    setTheme(theme: Theme) { ... },
  }),
}
```

Notice how now the `setTheme` doesn't have the `state` argument anymore, and this makes it the exact same function contract that we call in redux connect, making typing easier.

---

If we do
```ts
import selectorsPlugin, { select as rematchSelect } from '@rematch/select'
import * as models from '../models'

export type Reducers = {
  [key in keyof typeof models]: ReturnType<typeof models[key]['reducers']>
}

export type Selectors = {
  [key in keyof typeof models]: typeof models[key]['selectors']
}

export const select: Selectors = rematchSelect
```

And then do:

```ts
const mapState = (state: any) => ({
  theme: select.app.getThemeObject(state),
})

const mapDispatch = ({ app: { setTheme } }: Reducers) => ({
  setDarkTheme: () => setTheme('dark'),
})
```

Everything autocompletes wonderfully.

For that to work we just need the models api to support the `state =>` api in addition to the current one.

Thoughts?

brunolemos:
Closes #356

This PRs adds autocomplete for `mapState`, `mapDispatch` and `select`.

image

image


ShMcK:
This is absolutely amazing.
Just give me 24 hours to go over it, and ensure all is well before publishing :)

brunolemos:
@ShMcK thanks! Really appreciate the quick review+merge+release and the words.

PR #364, opened 2018-05-12
TypeScript improvements on store object

brunolemos:
Continuation of #357

Just a few more typescript improvements:

- Autocompletes store.dispatch
- Autocompletes store.getState() response
- Fixes store.subscribe return type
- Decrease the need of writing `typeof` everywhere
- Add link to counter example on TypeScript receipt

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -1,15 +1,20 @@
 # TypeScript
 
+### Examples
+
+- [Counter](../../examples/ts/count)
+
+### Changes
+
 Rematch can work with TypeScript with the following changes:
 
-##### Turn off "noImplicitThis".
+##### Turn off `noImplicitThis`.
 
 Rematch often specifies the context of `this`, leading to TS errors. 
 
 You can turn these off by specifying `noImplicitThis` as false in your "tsconfig.json".
 
-// tsconfig.json
-
+`tsconfig.json`
 ```json
 {
   "compilerOptions": {
@@ -17,6 +22,8 @@ You can turn these off by specifying `noImplicitThis` as false in your "tsconfig
   }
 }
 ```
+
+### Dependencies
 
 ##### Ensure `Redux@3.x` is not a dependency. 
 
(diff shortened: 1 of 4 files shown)


brunolemos:
Unrelated but not that much: rematch exports a `dispatch` out of nowhere, I don't think it's possible to add typings to that, and when it's not possible to add types it usually means something is wrong with the architecture.

Personally I don't think it makes sense to have that exported, `dispatch` should be only associated with a `store`, `store.dispatch`.

ShMcK:
The idea behind a global "dispatch" is that it solves a few issues:

##### Circular Dependency with using dispatch in a model

Model's cannot import `store.dispatch`, as the models are used to create the store.

There may be a better solution to this, love to hear your ideas and definitely open to any better solutions.

##### Multiple model dispatch

As we move towards supporting multiple stores, there are times when you may want to trigger changes across stores. Again, this is not highly recommended, and store.dispatch is preferable.

ShMcK:
Would it be possible to move the following into the selectors plugin?

```js
export { models };
export type models = typeof models;

export const select = _select as ExtractRematchSelectorsFromModels<models>;
const selectPlugin = _selectPlugin();
```

So that the select plugin could be configured with:

```js
const selectPlugin = _selectPlugin({
  models,
});
```

ShMcK:
This PR resolves a lot of issues with the selector plugin typings. 

However, I noticed that you cannot add Model generics to models with selectors.

```js
export const sharks: Model<SharksState> = {
  state: 0,
  selectors: {
    total(state: SharksState): number {
      return state;
    }
  }
};
```

Results in the following error:

```
incrementSharks: () => dispatch.sharks.increment(1),
// error TS2554: Expected 0 arguments, but got 1.
```

Removing the selectors from the model results in no warnings. Any ideas why this might be?

ShMcK:
I think as it stands, this is a significant improvement. I'll merge it in.

brunolemos:
Sorry, a bit busy, I can check your suggestions in a couple days

ShMcK:
Definitely understand. I've been trying to build upon your typings, but I think I need to gain a further understanding of newer TypeScript features. Great work :)

brunolemos:
@ShMcK not sure what we could do here... Maybe:

Instead of: `export const select = _select as ExtractRematchSelectorsFromModels<models>;`
How about: `export const select = getSelect<models>();`

Same thing for the global `dispatch`.

PR #369, opened 2018-05-14
Update misleading store import example

eric-burel:
Importing the store to access the dispatch method would create circular imports in real life case, we should import the `dispatch` function from rematch directly.


ShMcK:
Good catch. I've noticed some folks have been confused about this, hadn't realized it was due to the docs.

inDream:
But `dispatch` from core does not have auto complete feature made in #364

Issue #208, opened 2018-02-10
Promises returned from effects resolve to undefined


pierissimo:
Calling an effect return correctly a promise, but the value that will be resolved is always undefined.
Link to codesandbox:

https://codesandbox.io/s/3x3l27jk8p

ShMcK:
Which value are you expecting from the dispatch?

pierissimo:
the value returned from the effect. In the example provided, the string 'hey there'.

blairbodnar:
Ah, like this right?

```js
// 'count' model
effects: {
    async asyncIncrement() {
      await new Promise(resolve => {
        setTimeout(resolve, 1000);
      });
      dispatch.count.increment();
      return "hey there";
    }
  }
```

```js
// somewhere else
const handleAsyncIncrement = async () => {
    const value = await coiunt.dispatch.asyncIncrement();
    // expects value to be "hey there"    
  };
```

ShMcK:
Interesting idea. That could probably be added. Out of curiousity, could you give an example of a use case?

Initially I added promises to help ensuring dispatches happen in the correct order. For example, validate a form THEN route to a different page.

The thing about dispatches in Redux is that they don't return any value. You send a message out into the ether, and if it has a listener, something happens - otherwise nothing.

pierissimo:
Yes, the fact that they return a promise out of the box (without the necessity to add a custom middleware) is powerful.
But returning a promise that resolve to undefined, instead of a value could be misleading.
A use case could be the ability to fetch a value from the API, without saving it in the state (bypassing reducers/selectors).
I know that can be a bad approach, but there are situations where it can be useful.

The approach is used in vuex, where dispatched actions return promises that resolve to value.

ShMcK:
The idea of a "model" is to put all your related code in one place. 

Even if the code doesn't need to interact with Redux, I think that makes a lot of sense.

ShMcK:
The code that handles the returning promise is here: https://github.com/rematch/rematch/blob/master/src/plugins/effects.ts#L31-L37

Since it relies on Redux middleware, I'm not sure that it can return a value. But there's always a way.

pierissimo:
It's definitely possible with redux middlewares, as I'm using a custom middleware to return values after dispatching submit with redux-forms.

I'm not sure, but maybe the createDispatcher function should return a value:
https://github.com/rematch/rematch/blob/master/src/plugins/dispatch.ts#L17

pierissimo:
[...]

PR #371, opened 2018-05-14
fix(loadingEffectReturnValuePropagation): Propagate effect return value

eeynard:
Fix loading plugin regression regarding the issue #208

### DIFF
--- a/plugins/loading/src/index.ts
+++ b/plugins/loading/src/index.ts
@@ -118,12 +118,15 @@ export default (config: LoadingConfig = {}): Plugin => {
 
         // create function with pre & post loading calls
         const effectWrapper = async (...props) => {
+          let effetResult
+
           try {
             this.dispatch.loading.show({ name, action })
-            await origEffect(...props)
+            effetResult = await origEffect(...props)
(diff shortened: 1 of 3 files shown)


ShMcK:
Good catch, just one small change and should be ready to merge :)

eeynard:
Great ! Do you know when you will be able to merge and release the fix ?

ShMcK:
Sorry, I hadn't submitted the change request. But can merge right after :)

eeynard:
Oh yes you're right ! I just fixed it, sorry for the typo :)

eeynard:
@ShMcK I tried the v1.0.2 of the loading plugin and it does not seem to work. I checked the map files and it seems that the fix had not been packaged with the new release. Can't find the piece of code, any idea ?

ShMcK:
I must not have run build before publishing... my mistake.

eeynard:
It works with v1.0.3 ! Thanks

scientist1642:
Hmm, still Promise problem in `Loading 1.0.3` and `@rematch/core": "1.0.0-alpha.4"`, For me. 
If I remove Loading I can handle promises fine.

PR #373, opened 2018-05-14
chore: clone last 5 commits

DanielRuf:


ShMcK:
I'm not sure I understand. What is the benefit of limiting the git clone depth?

DanielRuf:
By default Travis clones the last 50 commits. Cloning 5 is sufficient and faster.

DanielRuf:
See https://docs.travis-ci.com/user/customizing-the-build#Git-Clone-Depthc

ShMcK:
Sorry to ask more, but what if a PR contains more than 5 commits?

DanielRuf:
The changes are still there in the later commits, just the history is truncated so you can not travel back in time more than these 5 commits. Git just has less to calculate. Also in general there is just one build per push, not for every commit.

Shorter log book / history = less diffs to look at.

https://git-scm.com/docs/git-clone#git-clone---depthltdepthgt

DanielRuf:
A PR with more than 5 commits still gets its build but history is truncated (from 50 to 5) and cloning / checkout is faster as there are less objects in your .git directory.

ShMcK:
That makes sense :)

PR #378, opened 2018-05-16
Enable "esModuleInterop" TypeScript setting

### LINKED ISSUE #349: TypeError: fe.default is not a function (@rematch/immer)
Versions:
    "@rematch/core": "^1.0.0-alpha.1",
    "@rematch/immer": "^0.1.0"

Errors:
```js
Uncaught (in promise) TypeError: fe.default is not a function
    at t.(:3000/anonymous function) (http://localhost:3000/static/js/bundle.js:896:11752)
    at rematch-immer.umd.js:1
    at computeNextEntry (<anonymous>:1:34000)
    at recomputeStates (<anonymous>:1:34300)
```

When I:
```js
import { init } from '@rematch/core';
import immerPlugin from '@rematch/immer';
import regex from './regex';

const immer = immerPlugin();

const store = init({
  models: {
    regex,
  },
  plugins: [immer],
});

export default store;


// regex.js
import uuid from 'uuid/v4';

export default {
  state: {
    ID: {
      example: [],
      regex: {},
      result: {},
    },
    byID: {},
  },
  reducers: {
    addExample(state, payload) {
      console.log('payload');
      
      const ID = uuid();
      state.ID.example.push(ID);
      state.byID[ID] = {
        text: payload,
        regex: [],
      };
      return state;
    },
  },
  effects: {},
};

// App.js
const mapDispatch = ({ regex: { addExample } }) => ({
  addExample,
});

@connect(() => ({}), mapDispatch)
export default class App extends Component {
...
  render() {
    console.log(this.props.addExample());
...
}
```

jackwilsdon:
Fixes #349.

See https://github.com/rematch/rematch/issues/349#issuecomment-389695117 for a better description of how this works.

I've run the unit tests against the module and it still seems to run fine.

I have not tested if this change affect the CommonJS and ES Module versions of the build, but it seems to fix the issue with the UMD version.

### DIFF
--- a/plugins/immer/tsconfig.json
+++ b/plugins/immer/tsconfig.json
@@ -6,7 +6,8 @@
     "lib": ["es2017"],
     "moduleResolution": "node",
     "declaration": true,
-    "declarationDir": "./dist"
+    "declarationDir": "./dist",
+    "esModuleInterop": true
   },
   "files": [
     "src/index.ts"


ShMcK:
published as v1.0.2 (next)

PR #379, opened 2018-05-18
[TypeScript] Add independent selector state type parameter to Model<S>

denisw:
Due to the `sliceState` feature of `@rematch/select`, the state passed to selectors might be different from the model state, so it needs to be possible to define the shape of the selector input state separately.

For this reason, this changes the `Model<S = any>` type to `Model<S = any, SS = S>` and defines the selectors as `(state: SS, ...args: any[]) => any`.

### DIFF
--- a/typings/rematch/index.d.ts
+++ b/typings/rematch/index.d.ts
@@ -114,7 +114,7 @@ export type ModelHook = (model: Model) => void
 
 export type Validation = [boolean | undefined, string]
 
-export interface Model<S = any> {
+export interface Model<S = any, SS = S> {
   name?: string,
   state: S,
   reducers?: ModelReducers<S>,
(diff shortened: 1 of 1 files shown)


ShMcK:
Looks good. Could you provide an example showing a needed sliceState typing? It may be useful for understanding and later documentation.

denisw:
@ShMcK Sure, how about this:

```typescript
import { Model, init } from '@rematch/core'

interface State {
  auth: AuthState
  chat: ChatState
}

interface AuthState {
  currentUserId: string | null
}

interface ChatState {
  messages: Array<{ senderId: string, text: string}>
}

const authModel: Model<AuthState, State> = {
   ...
}

const chatModel: Model<ChatState, State> = {
  name: 'chat',
  state: { messages: [] },

  selectors: {
    myMessages(state) {
      return state.messages.filter(m => m.sender.id === state.auth.currentUserId)
    }
  }
}

const models = {
  auth: authModel,
  chat: chatModel
}

const store = init({  
  models: {
    auth: authModel,
    chat: chatModel
  },
  plugins: [
    selectPlugin({
      sliceState: (state: State) => state
    })
  ],
})
```

ShMcK:
I see, in this case you use sliceState to provide the state typings.

I suppose it would also come in handy if using immutable.js.

Issue #361, opened 2018-05-10
Problem with types on alpha.4


wojciech-bilicki-oke:
```
// Type 'M[modelKey]["selectors"][reducerKey]' does not satisfy the constraint '(...args: any[]) => any'.
```
in `core/typings/index.d.ts`

This error occurs on the alpha.4 version. Everything is fine with alpha.2.

ShMcK:
Checkout the updated [ts example](https://github.com/rematch/rematch/tree/master/examples/ts/count): 

Some changes were made that allow for:
- autocompletion of dispatch methods
- validating of dispatch payload types

Love to hear what you think. Hopefully we can find a typing solution that makes everyone happy.

brunolemos:
@wojciech-bilicki-oke how does your reducer look like?

Also try this:

```diff
-export const app: Model = {
+export const app = {
```

Please remove the `:Model`.
Unfortunately this was the only way I found to have all the autocomplete possible by #357. If you know another way please let me know.

wojciech-bilicki-oke:
Hey, guys. Kinda busy right now, but I will give it a go tomorrow and let you know ;)

Valbrand:
Is `@rematch/select` somehow required? I'm getting the same error with `1.0.0-alpha.5`. I'll paste the code below, and let me know if there's something I can help with 😄 

```ts
import { init } from "@rematch/core";

const count = {
  state: 0,
  reducers: {
    increment(state: number) {
      return state + 1;
    }
  }
};

const store = init({
  models: { count }
});
```

wojciech-bilicki-oke:
Still the same with `1.0.0-alpha.5` even after removing the `:Model` from my models :(

alan-strohm:
Without `:Model` on my models, I get errors like 

> Property 'increment' does not exist on type { incrementAsync(payload: string): Promise<Void>; }.

Even with https://github.com/rematch/rematch/pull/382

alan-strohm:
I'm not sure if this is a good solution, but passing my model through a function like below seemed to do the trick of telling Typescript that my model extended Model without stripping it of the inferred type information.

```
function MakeModel<M extends Model>(m: M) {
  return m;
}
```

brunolemos:
@wojciech-bilicki-oke @Valbrand I'm not able to reproduce this typing issue. Is this code all that's required to reproduce it? It gives me no error.

@alan-strohm yes that could be nice, I can make a PR with this if @ShMcK agree it's a good idea

[...]

PR #391, opened 2018-05-28
TypeScript improvements - Part 3

brunolemos:
- [API] New `createModel` function for autocomplete purposes (https://github.com/rematch/rematch/issues/361#issuecomment-392510274)
- [API] New `getSelect` function for autocomplete purposes (https://github.com/rematch/rematch/pull/364#issuecomment-392596145)
- [API] New `getDispatch` function for autocomplete purposes (https://github.com/rematch/rematch/pull/369#issuecomment-391211890)
- [TYPINGS] Fix `RematchDispatch` `no compatible call signature` error
- [TYPINGS] Support `this` inside effects (alternative to #382)

I could not reproduce #361.

### DIFF
--- a/typings/rematch/index.d.ts
+++ b/typings/rematch/index.d.ts
@@ -44,29 +44,27 @@ export type ExtractRematchSelectorsFromModels<M extends Models, RootState = any>
   }
 }
 
-export type RematchDispatcher<P = any, M = any> = {
-  (action: Action<P, M>): Redux.Dispatch<Action<P, M>>;
-}
+export type RematchDispatcher<P = void, M = void> =
+  ((action: Action<P, M>) => Redux.Dispatch<Action<P, M>>)
   &
-    P extends void ? { (): Action<void, void>; } :
-    M extends void ? { (payload: P): Action<P, void>; } :
-    { (payload: P, meta: M): Action<P, M>; }
+  (P extends void ? () => Action<void, void> :
+    M extends void ? (payload: P) => Action<P, void> :
+    (payload: P, meta: M) => Action<P, M>)
 
-export type RematchDispatcherAsync<P = any, M = any> = {
-  (action: Action<P, M>): Promise<Redux.Dispatch<Action<P, M>>>;
-}
+export type RematchDispatcherAsync<P = void, M = void> =
+  ((action: Action<P, M>) => Promise<Redux.Dispatch<Action<P, M>>>)
   &
-    P extends void ? { (): Promise<Action<void, void>>; } :
-    M extends void ? { (payload?: P): Promise<Action<P, void>>; } :
-    { (payload?: P, meta?: M): Promise<Action<P, M>>; }
+  (P extends void ? () => Promise<Action<void, void>> :
+    M extends void ? (payload?: P) => Promise<Action<P, void>> :
+    (payload?: P, meta?: M) => Promise<Action<P, M>>)
 
 export type RematchDispatch<M extends Models | void = void> =
   (M extends Models
     ? ExtractRematchDispatchersFromModels<M>
     : {
-        [key: string]: {
-          [key:string]: RematchDispatcher | RematchDispatcherAsync;
-        }
+      [key: string]: {
+        [key: string]: RematchDispatcher | RematchDispatcherAsync;
+      }
     })
   & (RematchDispatcher | RematchDispatcherAsync)
   & (Redux.Dispatch<any>) // for library compatability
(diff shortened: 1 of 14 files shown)


ShMcK:
PR seems to be failing with:

```
typings/rematch/index.d.ts(123,48): error TS2526: A 'this' type is available only in a non-static member of a class or interface.
typings/rematch/index.d.ts(130,44): error TS2344: Type 'Self' does not satisfy the constraint 'Model<any, any, any>'.
```

I wonder if there is a way to type models without using `createModel`? I'm sure you've investigated this already, what are some of the issues?

brunolemos:
> I wonder if there is a way to type models without using createModel

I don't know any, unfortunately. When we do `const model: Model` we lose the inference. Maybe if there is some way to do something like `const model: extends Model` but I don’t think so.

> PR seems to be failing

I’ll check.

brunolemos:
@ShMcK here is an alternative based on a suggestion from @gtkatakura:

```ts

/////////////////////
// INTERNAL TYPINGS
/////////////////////
export interface Reducer { [key: string]: () => any }

interface Model<TReducers extends Reducer = Reducer> {
  name: string,
  reducers: TReducers,
}

/////////////////////
// USER MODEL
/////////////////////

const reducers = {
  setTheme: () => null,
}

export const model: Model<typeof reducers> = {
  reducers,
  name: 'abc',
}

/////////////////////
// TEST
/////////////////////

model.reducers.setTheme() // autocompletes
```

What do you think?

1. `model` would now autocomplete again
2. `reducers`, `effects`, etc, **declarations** would still not autocomplete
3. If they are wrong, the `model` line would complain

ShMcK:
I would favor a solution that resolves most use cases, even if features such as auto-completion become unavailable for the time being.

@brunolemos I only wish I had your advanced typing knowledge. The typings are quickly surpassing my understanding, which makes me worry that perhaps they might also become confusing for other developers as well.

The need for complex typings often indicates in itself a need to either:
- restructure the types
- alter the code base to avoid the need for complex typings

brunolemos:
Should I change anything in the PR?

ShMcK:
Got some time to go over this today.

I'm having trouble building this version with the following errors:

```
[!] (commonjs plugin) TypeError: parse is not a function in 
{path/to}/lib/index.js
TypeError: parse is not a function in 
{path/to}/lib/index.js
```

There is no `parse` inside of `lib/index.js`, so must be a commonjs issue. I'll look into it more. There don't seem to be any changes in this PR related to this issue.

ShMcK:
Resolved by rolling by locking commonjs plugin dependency. Must be an issue in the latest release.

typings/rematch/index.d.ts
```diff
@@ -119,7 +125,14 @@ export interface Model<S = any, SS = S> {
   state: S,
   reducers?: ModelReducers<S>,
   effects?: {
-    [key: string]: (payload: any, rootState: any) => void,
+    [key: string]: (
+      this: (
+        // ExtractRematchDispatchersFromModel<this> & // TODO: Make this work, if possible
+        { [key: string]: RematchDispatcher | RematchDispatcherAsync }
```
brunolemos: this.xxx access only reducers or effects too? If it's only reducers need to remove `| RemoteDispatcherAsync`

ShMcK: I don't believe there is a good reason to use `this` in a reducer, only effects. Otherwise it would lead to side effects.

brunolemos: No, I mean if we can only do this.reducerName or also this.otherEffectName

ShMcK: Effects should be supported as well in the typings.

ShMcK:
There seems to be an issue with effects & this typings in the example:

```
examples/ts/count/src/models/sharks.ts
(16,22): Argument of type 'number' is not assignable to parameter of type 'Action<void, void>'.
```

ShMcK:
Will merge with one additional fix to be added after. The select plugin throws a few warnings due to "ambient declarations must be string". 

I believe the plugin needs some work, and can be ignore for now using `@ts-ignore`.

ShMcK:
published as @rematch/core@1.0.0alpha.7

brunolemos:
Thanks! Not sure if you noticed that my `this` type implementation is different than the one you merged, you think that one is more correct?

ShMcK:
I believe this one is more correct.

`this` = `dispatch[modelName]`, so it can only be a function called with (action, meta).

Whereas `RematchDispatch` is both a Redux.dispatch function and a function with keys on it.

But if I'm wrong, it wouldn't be the first time :)

PR #401, opened 2018-06-09
Fix @remach/loading build 

jancassio:
The current build process [@remach/loading](https://github.com/rematch/rematch/tree/master/plugins/loading) is divided in two parts

1. typescript compiler generate output and type definitions
2. rollup generate compatible modules from generated output

I moved everything into a single rollup process, now both definitions and `.js` files are generated in a single build.

Also, I've added into `packages.json` the property `types` to help TypeScript to find definitions into the module.


ShMcK:
Thanks :) I have a good understanding of how frustrating it can be to put fixes in for rollup/typescript build processes.

I had issues running build on this branch, but after switching to "rollup-plugin-typescript2" everything is working. Will merge fixes on top of this branch once merged and reference them here.

PR #407, opened 2018-06-11
Update redux.ts

RubyLouvre:
add isEmptyObject method Quickly determine whether an object is empty

### DIFF
--- a/src/redux.ts
+++ b/src/redux.ts
@@ -26,7 +26,7 @@ export default function({
 	this.mergeReducers = (nextReducers: R.ModelReducers = {}) => {
 		// merge new reducers with existing reducers
 		this.reducers = { ...this.reducers, ...nextReducers }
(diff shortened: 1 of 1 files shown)


ShMcK:
It seems the TS linter is unhappy with the use of a `for in` loop, in this case. 

The following utility function should work:

```js
function isEmptyObject(obj?: object): boolean {
	return Object.keys(obj || {}).length === 0
}
```

It would be great too if you added it to the `utils` folder as an export. Should clean things up a bit.

RubyLouvre:
`for in ` ’s performance is higher, as long as a property is detected。You can't ignore performance just because it looks good

jackwilsdon:
Whilst `for in`'s performance is about 50% better than `Object.keys`, `Object.keys` can still do 14,998,528ops/sec on my i5-5250U which I think is more than enough.

PR #411, opened 2018-06-11
fix NODE_ENV in build script

saintwinkle:
`process.env.NODE_ENV` remains `undefined` in former situation.

This PR should fix the issue, and can be built to minified dist.

---
Love `rematch`, thanks for creating this framework.


ShMcK:
Thanks, nice fix :)

PR #412, opened 2018-06-11
Strict null check

### LINKED ISSUE #406: failed to compile use typescript
ts version: 2.9.1 
rematch version:  1.0.0-alpha.8

get this error: 
image

brunolemos:
Fix #406

@ShMcK can you finish this PR? It requires more knowledge about the internals.
Please run `npx tsc -p .` to see the remaining errors.

They are mostly because the `name` field it optional, so things like `this.reducers[model.name]` could cause an issue (`this.reducers[undefined]`). But can they really be undefined? Maybe the `name` field should be required instead of optional? Or are you setting the name automatically somewhere else?

> If you are pretty sure they will never be undefined for some reason, you can just add an `!`, e.g. `this.reducers[model.name!]`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,9 +14,9 @@ const dispatches = {}
  * @param action
  */
 export const dispatch = (action: R.Action) => {
-  for (const storeName of Object.keys(stores)) {
-    stores[storeName].dispatch(action)
-  }
+	for (const storeName of Object.keys(stores)) {
+		stores[storeName].dispatch(action)
+	}
 }
 
 /**
@@ -26,11 +26,11 @@ export const dispatch = (action: R.Action) => {
  * returns an object with key: storeName, value: store.getState()
  */
 export const getState = () => {
-  const state = {}
-  for (const name of Object.keys(stores)) {
-    state[name] = stores[name].getState()
-  }
-  return state
+	const state = {}
+	for (const name of Object.keys(stores)) {
+		state[name] = stores[name].getState()
+	}
+	return state
 }
 
 /**
@@ -41,7 +41,7 @@ export const getState = () => {
  * returns the dispatch object with typings information
  */
 export function getDispatch<M extends R.Models>() {
-  return dispatch as R.RematchDispatch<M>
+	return dispatch as R.RematchDispatch<M>
 }
 
 /**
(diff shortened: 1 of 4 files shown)


ShMcK:
# Internal & External Typings

Currently there are two sets of typings for config. Users may omit information, but internally it gets filled in. See `Config` as an example.

```ts
// user set config
export interface InitConfig<M extends Models = Models> {
  name?: string,
  models?: M,
  plugins?: Plugin[],
  redux?: InitConfigRedux,
}

// how models are used internally with defaults all set
export interface Config<M extends Models = Models> {
  name: string,
  models: M,
  plugins: Plugin[],
  redux: ConfigRedux,
}
```
# Models

I think these model typings errors might be cleaned up if the same were done as Config: provide an external and internal set of typings.

### model.name

Currently, `model.name` does not need to be defined within the config, but it becomes defined internally. See the following types and where they are found in the source code:

### reducers

The same goes for `reducers`. A user can set no reducers, but internally reducers would just be set to an empty object.

--- 

@brunolemos What are your thoughts?

ShMcK:
I'll create a `ModelConfig` & `Model` typings, to differentiate between the two.

brunolemos:
Ok, you can add the `!` or create new types e.g. 

```ts
export interface InternalConfig extends Config {
  name: string,
}
```

ShMcK:
I've put some efforts onto a [branch](https://github.com/rematch/rematch/tree/brunolemos-strict-null-check), resulting in errors related to Config typings.

```
const config: R.Config = mergeConfig({ ...initConfig, name })
```

I wish I had a better understanding of complex typings.

brunolemos:
@ShMcK I think you can push to my branch because `Allow edits from maintainers.` is marked.

Pushed here, it should be all right now.

ShMcK:
@brunolemos I made a PR on top of your branch in your forked Rematch repo. Unfortunately, I haven't been able to find a way to push directly to a PR forked and living in another repo.

brunolemos:
@ShMcK not sure if I understand what you are trying to do with the branch, isn't it the same code of this PR here? #412 is ready for merge

andrewheekin:
good to merge here?

ShMcK:
Yeah, I'm not what I was trying to do with the typings either :)

No longer getting errors from tsc compilation. Will merge and release tonight.

Issue #400, opened 2018-06-09
Accessing other model selectors in a selector


d3dc:
Say I have a normalized state and a selector to denormalize it. Would it be considered a bad pattern to stack these ~~like I do with reselect~~? 

It seems to me like selectors should have the same api as effects:

```js
{
  hydrate: (state, input) {
    return denormalize(input, schema, state)
  },
  getBook(state, id) {
    return this.hydrate(id)
  },
  bookList(state) {
    return this.hydrate(Object.keys(state.books))
  },
  sortedBookList(state) {
     return sortBy(this.bookList(), 'modified')
  }
}
```

And while we're at it, could concerns around accessing external state like #270 be addressed by the DI API you are working on for effects? ~~I don't see a  reason for this other than api parity.~~
```js
  selectors: (select) => ({
    schemaSelector(state) {
      return select.storage.denormalize(state.list, schema)
    }
  })
```

d3dc:
I was thinking having `this` bound lets store model state be built up from smaller pieces. I can decorate a model to provide a default implementation of `hydrate` (and `dehydrate`) and then rely on `this.hydrate` in the specific domain selectors.

But I guess you could also just have a single `storage` model that you select and dispatch to - just like a database. And either I'm forgetting why this is a bad idea, or that seems like nice decoupling. I can have multiple schemas relating to each other and each entity type can still have very unrelated sagas.

d3dc:
[I went ahead and mocked this up here](https://codesandbox.io/s/ql6m574l84)

I'm running into an issue with the `this` binding in selectors and may have just missed something.

ShMcK:
There are currently some issues with the selectors plugin & 1.0.0-alpha. We'd be happy to take in any PRs to fix issues. 

Unfortunately, there are a lot of different people with different use cases and ideas for the selectors plugin. I've found it's just easier to write your own selectors wrapper and use it where needed.

d3dc:
@ShMcK I can understand there's a lot of issues still around with 1.0 looming (I've had to dig a bit to get answers to some of my questions).

My intent is definitely to put in a PR eventually - just trying to spitball first. I'll definitely take a look at #386, #360, and #309. Am I missing any? It doesn't seem like there's any large blockers to what I'm doing.

What would it take to PR an API change to the selector plugin? I've gotten my example working and added a little more state to illustrate being dependent on another reducer - it feels natural to write my outputs the same as my inputs to the state and I don't see how this would interfere with reselect use or anything.

[...]

PR #399, opened 2018-06-08
Proposal: effects as fn, with local dispatch passed in

ShMcK:
### Proposal: optionally pass local `dispatch` into effects.

```js
const model = {
  effects: (dispatch) => ({ ... })
}
```

TODO
- [ ] fix typings

PR #413, opened 2018-06-12
Proposal: Select is a mirror of Effects

d3dc:
**Proposal: optionally pass local select into selectors and allow directly calling other selectors on the same model.**

```js
{
  selectors: (select) => ({
    double: s => s * 2,
    quadruple: s => select.model.double() * 2
  })
}
```

[I recreated my external selectors inside rematch and it seems pretty clean](https://codesandbox.io/s/ql6m574l84)

**TODO:**
- [x] make multiple-store-safe
- [ ] typings
- [ ] examples
- [ ] validate - is this a good idea???

---

**Motivation**
Following #399, I made some comments in #400.

There are an alarming number of issues related to people using the global `dispatch` (not so much `select`, but it embraces the global aspect). Most of the outstanding issues I could find could be solved by using dependency injection through `connect`. It seems like that API is really what everything should look like.

This PR binds selectors to the current store and lets them call each other without passing down state - Exactly like dispatchers. So you have  state changes with no dependencies and identical dependent storage and selection flows. (pure selection is just consuming raw state, no?)

Unlike dispatchers, to establish bindings, the selectors are stored in the state and retrieved when the global `select` is used. I had a few ideas after reading #270, and it sounded like this might be the part that was objectionable...

### DIFF
--- a/plugins/select/package.json
+++ b/plugins/select/package.json
@@ -32,7 +32,7 @@
   },
   "browser": "dist/rematch-select.umd.js",
   "devDependencies": {
-    "@rematch/core": "^1.0.0-alpha.8",
+    "@rematch/core": "^1.0.0-alpha.9",
     "rollup": "^0.60.1",
     "rollup-plugin-commonjs": "^9.1.3",
     "rollup-plugin-replace": "^2.0.0",
@@ -42,7 +42,7 @@
     "uglify-es": "^3.3.9"
   },
   "peerDependencies": {
-    "@rematch/core": ">=1.0.0-alpha.8"
+    "@rematch/core": ">=1.0.0-alpha.9"
   },
   "authors": [
     "Blair Bodnar <blairbodnar@gmail.com> (https://github.com/blairbodnar)",
--- a/plugins/select/package.json
+++ b/plugins/select/package.json
@@ -32,7 +32,7 @@
   },
   "browser": "dist/rematch-select.umd.js",
   "devDependencies": {
-    "@rematch/core": "^1.0.0-alpha.9",
+    "@rematch/core": "^1.0.0-beta.1",
     "rollup": "^0.60.1",
     "rollup-plugin-commonjs": "^9.1.3",
     "rollup-plugin-replace": "^2.0.0",
@@ -42,7 +42,7 @@
     "uglify-es": "^3.3.9"
   },
   "peerDependencies": {
-    "@rematch/core": ">=1.0.0-alpha.9"
+    "@rematch/core": ">=1.0.0-beta.1"
   },
   "authors": [
     "Blair Bodnar <blairbodnar@gmail.com> (https://github.com/blairbodnar)",
--- a/plugins/select/src/index.ts
+++ b/plugins/select/src/index.ts
@@ -1,20 +1,30 @@
-import { ExtractRematchSelectorsFromModels, Model, Models, Plugin } from '@rematch/core'
+import { ExtractRematchSelectorsFromModels, Store, Model, Models, Plugin } from '@rematch/core'
 
 export const select = {}
+
+const stores = []
+const localSelects = []
+
+const mapStateToStore = (state: any) =>
+	stores.findIndex((store) => store.getState() === state)
+
+export function withSelect(mapSelectToProps: Function) {
+	return (state, ...args) =>
+		mapSelectToProps(
+			localSelects[mapStateToStore(state)],
+			...args
+		)
+}
 
 export function getSelect<M extends Models = Models>() {
 	return select as ExtractRematchSelectorsFromModels<M>
 }
 
 export interface SelectConfig {
-  name?: string,
 	sliceState?: any,
 }
 
 const validateConfig = (config) => {
-  if (config.name && typeof config.name !== 'string') {
-    throw new Error('select plugin config name must be a string')
-  }
   if (config.sliceState && typeof config.sliceState !== 'function') {
     throw new Error('select plugin config sliceState must be a function')
   }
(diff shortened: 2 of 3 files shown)


d3dc:
With how I handle multiple stores I guess this becomes a little more opinionated (I copy pasted the loading plugin tho). Some short examples around what type of style this coerces:

```js
@connect(
  ({ select, model }) => ({
    aValue: select.model.value(),
    pureA: model.a
  }),
  (dispatch) => ({
    anAction: dispatch.model.action
  })
```

```js
{
  reducer: {
    pureChangeA: (state, a) => { ...state, a },
    pureChangeB: (state, b) => { ...state, b }
  },
  effects: (dispatch) => ({
    dependentChange() {
      this.pureChangeA(1)
      this.pureChangeB(2)
      dispatch.external.change(3)
    }
  }),
  selectors: (select) => ({
    dependentValue(state) {
      return state.a + select.external.a
    }
  })
}
```

d3dc:
@ShMcK I'm beginning to think I have to be typing for my brain to work.

That state key is a config option as well. You can pass a `SelectConfig` to the plugin (`createSelectPlugin` now), just like the loading plugin:
```js
SelectConfig {
  name?: string,
  sliceState?: any,
 }
```

ShMcK:
Definitely keep the select key name configurable.

I'd say play around with the plugin for a few days, and see if you're happy with it. It seems to have some good developer friendly ideas in it so far. And we are in need of a Select plugin replacement.

ShMcK:
One thing to keep in mind. It's definitely not best practices to store functions in Redux state:

https://stackoverflow.com/questions/35325195/should-i-store-function-references-in-redux-store

d3dc:
Good call on that. You might have to avoid rehydrating the `select` key. This should work for the most part by registering the selectors with a deterministic ref. The thing I don't know exactly is: What is a deterministic identifier for a store? It seems like so far its just instantiation order...

ShMcK:
Stores may be named with:

```js
init({
  name: 'someName'
})
```

Otherwise the index as a string is used.

d3dc:
I'll look into using the index I guess. 

I think its valuable to have the model semantics line up, so there's still a point, but `connect`ing `select` doesn't seem like it can work. 

You could:
```js
import { session } from 'select'

connect((state) => {
  const select = session(state)

  return {
    aValue: select.model.value()
  }
})
```

But, oh boy, does the idiot in my brain already have questions about what a session is.

d3dc:
@ShMcK I'm still stuck at finding the store name. Its never passed to plugins directly.

d3dc:
Regarding Vuex: I think `selectors` and Getters line up in my mental model - but I don't see the reason to separate them from the state they consume. If I'm registering a model, I want to also register common ways to access its data, no?

The one advantage to providing your own selectors that I think a library should tackle is the automatic wiring including: state slicing and calling other "getters" since these are managed contexts.

Computed Properties exist on the Vue - watching the dependencies etc. is the job of something like `reselect` here I would think. It could be interesting to do `connect(...computed())` for that property like syntax, however I think thats a different plugin.

ShMcK:
@d3dc that sounds like a bug. Plugins should have access to all stores.

As a hacky way: its possible to use `onStoreCreated` to collect all stores.

I believe these stores all have a "name" as it is used in the devtools (with multiple stores you can switch between them by name).

ShMcK:
I guess computed properties are just a way to calculate new state keys.

```js
state: { a: 1, b: 2},
computed(state) {
  total() {
     return state.a + state.b
  }
}

connect(state => ({
  total: state.total
})
```

These would not be included in the state. 

However, I wouldn't want computed properties to necessarily run every time the state updates. Seems heavy.

d3dc:
@ShMcK I think the answer is: If you don't want to run a selector a lot, wrap it in `createSelector` from `reselect`.

In models, this breaks using `this`. In connect, you would have to create your computed property config ahead of time.

d3dc:
So, pending beta 2, this should work fully with the `withSelect` decorator I mentioned (in addition to imported `select` binding to the state its passed).

I store a ref in the state to quickly lookup the local select. 
That ref happens to be the store name (which doesn't exist yet).

@ShMcK I originally started writing exposing the store ref as its own plugin, but decided to stop. Do you think this is something that needs to be extracted now? At the very least I can bring back its tests.

d3dc:
🤔 I see your point now. State slicing prevents reselect from re-running selectors that depend on other selectors. Vuex Getters however, always update.

The intuition to provide your own selectors is exactly the "Computed properties attached to the Vue" part... Does having view specific selectors in a global registry feel worthwhile?

And semantically, does what I've been writing fall under a different namespace? `select` isn't exactly `createSelector`

ShMcK:
Sorry my mind was heading in other directions today. I'll try to get back to you with something more thoughtful tomorrow. 

I appreciate the time and effort you've put into trying to get the select plugin right :)

d3dc:
No worries, I think I might actually close and break this PR apart now. It really is a mess.

I think whats worth taking away:

1. dependency injection should be encouraged. Repeating it throughout the API reinforces it and keeps things consistent.

2. The `select` plugin right now exists as a way to inject the slice.

3. The "Getters" in this PR are difficult to use, since they don't track dependencies.

---

For `select` would it make sense to just use it as wiring to inject things?

```js
{
  state: { a: 1 }
  selectors: (selector, slice) => ({
    valueA: selector(state => state.model.a)
    valueB: selector(slice, state => state.a)
    valueC: slice(slice => slice.a)
  })
}
```

---

Getters seems like a potential foot gun if you put expensive operations in them and they don't have caching. I think this could be solved with a very small utility that mocks running the getter since getters follow a very strict lexical structure.

```js
store.model({
  selectors: (selector) => {
    return computed.reduce((acc, compute) => ({
      [compute.name]: selector(createDepsSlice(compute.getter), compute.getter)
    }))
  }
})
```

PR #415, opened 2018-06-12
[WIP] Upgrade react-navigation-plugin to react-navigation v2

bosung90:
This is an upgrade to react-navigation plugin to support react-navigation v2.
This is WIP because currently I could not get dispatch.nav.navigate to work.
Also starting rematch v1.0 models require state therefore example to add getCurrentRouteName selector does not work anymore.


ShMcK:
Expo is now working with RN v2. I'll also be upgrading.

Is this PR ready?

bosung90:
No, apparently react-navigation decided to change the way redux integration works so this plugin will no longer with latest version of `react-navigation-redux-helpers`

derrickbeining:
@bosung90 , what needs to be done for this to be ready to merge? Happy to pitch in if you like.

I was excited to start migrating my client's React Native project from vanilla `redux` to `rematch`, but got snagged on this issue because we're using React Navigation v2. I'd love avoid having your use your work around for very long.

ShMcK:
Close in favor of #447.

PR #416, opened 2018-06-12
Update devtools.md to support react-native

bosung90:
Add guide to support redux devtools for react-native


docs/recipes/devtools.md
```diff
@@ -18,6 +18,39 @@ init({
 })
 ```
 
+### Redux-Devtools-Remote
```
ShMcK: Is it Remote-Redux-Devtools ?

PR #422, opened 2018-06-14
PluginFactory exposes config.

d3dc:
Rematch initializes PluginFactory with a config

### DIFF
--- a/src/pluginFactory.ts
+++ b/src/pluginFactory.ts
@@ -6,7 +6,7 @@ import validate from './utils/validate'
  *
  * makes Plugin objects extend and inherit from a root PluginFactory
  */
-export default (config) => ({
+export default (config: any) => ({
 	config,
 
 	/**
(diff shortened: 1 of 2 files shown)


src/pluginFactory.ts
```diff
@@ -6,7 +6,9 @@ import validate from './utils/validate'
  *
  * makes Plugin objects extend and inherit from a root PluginFactory
  */
-export default () => ({
+export default (config: any) => ({
```
ShMcK: In this case the type would be `R.Config`

ShMcK:
Unfortunately, this PR is breaking TypeScript in some horrible ways that will take me a while to wrap my head around :) PluginFactory inherits from Plugin.

d3dc:
... what if it was just `rematchConfig`? The typings would be a little
confusing.

For my use, the store name PR is working.

ShMcK:
TypeScript and I have a love hate relationship. I'm sorry you had to come in the middle of this. We'll sort it out, we also do.

PR #428, opened 2018-06-15
Select plugin can also wire dependencies.

d3dc:
I typed a lot about the importance of dependency injection and getting data out of the state in #413.

The advantage I see to the current recommendation of "import your own selectors" is that you can declare inputs to selectors much more directly. All the select plugin currently does is remove control over that from you.

This PR seeks to make it more natural to handle the entire API of `reselect` inside the models.

First off, it imports `reselect` directly, and lets you override this through configuration. If we're imitating the `reselect` api, why not just be the `reselect` api?

Models are auto-wired with dependencies, completely decoupling them for testing:
```js
{
  state: 1,
  selectors: (slice, createSelector, hasProps) => ({
    valueA () {
      return slice(a => a)
    },
    valueB () {
      return createSelector(slice, a => a)
    },
    valueC (models) {
      return createSelector(slice, models.other.b, (a, b) => a + b)
    },
    valueD: hasProps(function(models, props) {
      return createSelector(slice, a => a + props))
    })
  })
}
```

Since the plugin is built around the `reselect` api, why not make it the default to get a little more performance and change `select`:
```js
connect(select(models => ({
  x: models.one.a,
  y: models.two.b
})))
```

This makes a [structuredSelector](https://github.com/reduxjs/reselect#createstructuredselectorinputselectors-selectorcreator--createselector) or you can configure it too.

**Todo:**
- [x] tests
- [x] docs
- [x] examples

### DIFF
--- a/plugins/select/src/index.ts
+++ b/plugins/select/src/index.ts
@@ -1,44 +1,125 @@
-import { ExtractRematchSelectorsFromModels, Model, Models, Plugin } from '@rematch/core'
+import { Store, Model, Plugin } from '@rematch/core'
+import { createSelector, createStructuredSelector } from 'reselect'
+export { createSelector, createStructuredSelector } from 'reselect'
 
-export const select = {}
+export const selectors = {}
 
-export function getSelect<M extends Models = Models>() {
-	return select as ExtractRematchSelectorsFromModels<M>
+/**
+ * Maps models to structured selector
+ * @param  mapModelsToSelectors function that gets passed `selectors` and returns an object
+ * @param  structuredSelectorCreator=createStructuredSelector if you need to provide your own implementation
+ *
+ * @return the result of calling `structuredSelectorCreator` with the new selectors
+ */
+export function select(
+	mapModelsToSelectors: any,
+	structuredSelectorCreator = createStructuredSelector
+) {
+	let func = (state, props) => {
+		func = structuredSelectorCreator(mapModelsToSelectors(selectors))
+		return func(state, props)
+	}
+
+	return (state, props) => func(state, props)
 }
 
 export interface SelectConfig {
-	sliceState?: any,
+	sliceState?: any;
+	selectorCreator?: any;
 }
 
-const selectPlugin = ({
-	sliceState = (rootState, model) => rootState[model.name],
-}: SelectConfig = {}): Plugin => ({
-	exposed: { select },
-	onInit() {
-		this.validate([
-			[
-					typeof sliceState !== 'function',
-					`The selectPlugin's getState config must be a function. Instead got type ${typeof sliceState}.`,
-				],
-			])
-	},
-	onModel(model: Model) {
-		select[model.name] = {}
+const validateConfig = config => {
+	if (config.sliceState && typeof config.sliceState !== 'function') {
+		throw new Error('select plugin config sliceState must be a function')
+	}
+	if (config.selectorCreator && typeof config.selectorCreator !== 'function') {
+		throw new Error('select plugin config selectorCreator must be a function')
+	}
+}
 
-		Object.keys(model.selectors || {}).forEach((selectorName: string) => {
-			this.validate([
-				[
-					typeof model.selectors[selectorName] !== 'function',
-					`Selector (${model.name}/${selectorName}) must be a function`,
-				],
-			])
-			select[model.name][selectorName] = (state: any, ...args) =>
-				model.selectors[selectorName](
-					sliceState(state, model),
-					...args,
-				)
-		})
-	},
-})
+const createSelectPlugin = (config: SelectConfig = {}): Plugin => {
+	validateConfig(config)
 
-export default selectPlugin
+	const sliceState = config.sliceState || ((state, model) => state[model.name])
+	const selectorCreator = config.selectorCreator || createSelector
+
+	const slice = model => stateOrNext => {
+		if (typeof stateOrNext === 'function') {
+			return selectorCreator(state => sliceState(state, model), stateOrNext)
+		}
+		return sliceState(stateOrNext, model)
+	}
+
+	const hasProps = inner =>
+		function(models) {
+			return selectorCreator(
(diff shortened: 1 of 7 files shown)


ShMcK:
Does the plugin always require `reselect` in order to work, or might there be situations where `reselect` is not needed?

Perhaps "slice" should be the first param?

d3dc:
@ShMcK This does not require reselect to work, not providing `createSelector` will cause selector to just rewrap its last parameter.

One of the things I really missed is that `inputs` is array-like in reselect and passed in as the args to the function. (No, I don't know how I missed it)

I'm pretty sure that the "no reselect" case is wrong because it doesn't duplicate this behavior.

The other reason I bring up the `inputs` array, is that it really does give you a place to put your dependencies. I've spent a little bit trying to track the dependencies now, and while it works, its often just much more obvious and exhaustive to specify your dependencies when you call `createSelector`.

With that slightly renewed intuition, I'm wondering if my previous PR was a better base for explicitly declaring dependencies. Something ala:
```js
selectors: (select, slice, createSelector) => ({
  expensive: createSelector(
    [slice, select.other.calculated, select.some.bigValue],
    (state, calculated, bigValue) => doABigOperation(calculated, bigValue) + state.offset
})
```

williamconnolly:
I don't have anything to contribute - damn full time work -- but wanted to let you know I really appreciate the work being done here. Had to take a break from my React/Rematch Chrome Extension because I was just rewriting selectors over and over

d3dc:
So this lets you write very testable models, but what about consuming the selectors?

I think the `select` that you import should be completely different:
```js
select(models => ({
  x: models.one.a,
  y: models.two.b
}))
```

Where this calls a [createStructuredSelector](https://github.com/reduxjs/reselect#createstructuredselectorinputselectors-selectorcreator--createselector) equivalent.

ShMcK:
Let me know when you feel the select plugin PR is ready for review :)

d3dc:
Seems to pass my "complicated app" test at:
https://codesandbox.io/s/ql6m574l84

Its about time to set this free.

d3dc:
@ShMcK Perhaps not obvious, but the point is `todos` creates a new person, but its not linked to the todo.

Since I didn't want to handle back refs or anything to search for things that shouldn't be deleted, I added a button.

d3dc:
@ShMcK I cleaned up the normalizr stuff a little better.

plugins/select/src/index.ts
```diff
@@ -1,44 +1,127 @@
...
+	const slice = model => stateOrNext => {
+		if (typeof stateOrNext === 'function') {
+			return selectorCreator(state => sliceState(state, model), stateOrNext)
+		}
+		return sliceState(stateOrNext, model)
+	}
+
+	const hasProps = inner =>
+		function(models) {
+			const forProps = selectorCreator(
```
d3dc: This probably needs to be _in_ one more scope

ShMcK:
I notice the `selectors` seem to be generated on every function call.

In this case, `createSelector` is called repeatedly. 

```js
selectors: (slice, createSelector) => ({
    todoIds({ storage }) {
      return createSelector(
        state => storage.rawCollection(state, Todo),
        collection => Object.keys(collection)
      );
    }
})
```

Might there be a way to create a selector once?

ShMcK:
Another idea. It may be more ideal if `select` were bound to a store.

```js
export const { dispatch, getState, select } = init()
```

This would avoid the issue of a global select with multiple stores.

d3dc:
@ShMcK Selector factories are ran once. I think its like the first line of "Basics" and then I never mention it again. A factory is only run once all of its dependencies are there (which is why is declares them)

I've actually been re-exporting `select` alongside my store anyway. A way for plugins to add this would be great.

ShMcK:
Excellent. As you can tell, I'm a "skimmer". I hadn't realized the selectors function would run once as a factory. That's great.

I agree, it would be great if a plugin could be returned from `init`, and select would be a great use case. I'll make it happen.

d3dc:
@ShMcK Worth highlighting on factories: The third prop `hasProps` creates a memoized factory. It **will** re-run its factory whenever its called with a different payload.

d3dc:
@ShMcK I rebased this and added a dependency on the github repo to test #443. 

I'm pretty sure I have no clue how to write the typings for `select` or how to put them on `RematchStore`.

ShMcK:
I could guess that from the e779910 commit message. ;)

I'll have a look and see when it's ready. There should be a way to autocomplete select, as in dispatch.

ShMcK:
@brunolemos, if this is something you're interested in, the select plugin typings seem to be quite similar to dispatch/effect typings.

d3dc:
Here's a fun monkey wrench: To use Immutable.js or other custom `sliceState` functions for _other_ model dependencies, you need to access to every model's slice function.

I'm currently solving this with a method on `slice`:
> slice.for(modelName: string) =>  sliceFn | ((state) => undefined)

This works and makes it obvious that slices may not have arrived yet. It is different semantics than `select` - The plugin waits for all synchronous dependencies before injecting `select` and will throw an error for async dependencies that aren't there when `store.model` is called.

Should slice be injected with the same semantic? "These are the currently registered slices, no future values"

ShMcK:
Would you consider the select plugin "fully baked" by now?

d3dc:
@ShMck Using it with ~20ish models with no hiccups yet. There's probably already too much here. 

👍

ShMcK:
Alright. This is basically my last major barrier to 1.0:
- a select plugin that people are happy with (if that's possible).

If you can give me a few days to play around with it first, that'd be great. Really sorry for the delay.

ShMcK:
20 days later... honestly, great work.

I'll merge, with the caveat that it won't take long before TypeScript users begin to complain. My TS linter is displaying a few fixable warnings. 

Regardless, TS typings are not working with the previous version of select. See #474.

Will merge and publish as 2.0.0 within a day.

PR #429, opened 2018-06-15
[WIP][Proposal][Plugin] Store name plugin

d3dc:
This PR adds a plugin that extracts the common logic for giving each store's state a unique ref - the store name. You can use this ref to store and retrieve any non-serializable data your plugin needs.

stores the current store's name in it's state under a configured `name` key or `STORE_NAME_KEY = '@@rematchStoreName'`

plugins can retrieve the key from `this.storeNameKey`

### DIFF
--- a/plugins/storeName/.npmignore
+++ b/plugins/storeName/.npmignore
@@ -1,0 +1,10 @@
+# directories
+lib
+src
+test
+examples
+
+# files
+.gitignore
+.prettierrc
+yarn.lock
--- a/plugins/storeName/README.md
+++ b/plugins/storeName/README.md
@@ -1,0 +1,30 @@
+# Updated
+
+Rematch plugin for keeping a store's name in state for future lookup.
+
+This is primarily used for other plugins.
+
+
+### Install
+
+```
+npm install @rematch/exposeStoreName
+```
+
+### Setup
+
+```js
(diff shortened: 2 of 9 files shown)


ShMcK:
Looks like it should be able to without needing to wrap it in an object. I'd call that a bug.

I have some thoughts on why this might be useful - what kind of use cases are you considering?

Is there a good argument that this might be better moved to core?

d3dc:
Oof. This is WIP because `this.storeName.key` is awkward.

It might also be WIP because nothing uses it yet.

ShMcK:
Heads up on changes in beta.2. See the [changelog](https://github.com/rematch/rematch/blob/master/CHANGELOG.md).

Includes `plugin.config` & the ability to "name" a store on init.

d3dc:
So, coming back to this, this keeps the structure of your data outside of state as well. I think most cases would be best solved with that in the store. For my use case, getters would be still be external to the store, but the plugin would just track a unique id for each function - the shape of the structure indicating things like what models had registered etc. would be in the store.

The main benefit then to this plugin is the single-point of contact.

d3dc:
I don't have a great use case for this yet,

With the current addition of adding store properties, there might be potential to retrieve those additional store properties in `connect` without imports. But that might also be better with models insulating it.

PR #433, opened 2018-06-16
[typings] Fix getState autocomplete

brunolemos:


ShMcK:
I was wondering about this. 

How is it possible to type multiple stores which each taken different models?

brunolemos:
How would the js api be like?

ShMcK:
`getState` returns an object of stores keyed on store names.

```js
import { init, getState } from '@rematch/core'

const store1 = { name: 'store1', models }
const store2 = { name: 'store2', models: otherModels }

getState()
/*
 { 'store1': { count: 1 }, 'store2': { otherCount: 2 } }
*/
```

brunolemos:
getState is a confusing name for this because on redux it is for one store.

Well, it is possible but would require more changes. You could do something like `const stores = { store1, store2 }` but we would still need a way to get the models from the store. E.g. store1.models

ShMcK:
You're absolutely right. 

1. `getState` leads to confusion. `getStates` may be a better name. This does not solve the typings problem.

2. It may be possible to use `getState` if there is a single store, and throw a warning recommending to use `getStates` if the user has multiple stores. Most people only use one store, and it would avoid a breaking change with 1.0.0.

brunolemos:
So just to summarize:

> Q: Is it possible to type multiple stores which each taken different models?

> A: Well, it is possible but would require more changes.

What I mean is that it would require some api changes. One suggestion was:

> getStates(stores)

^ Focus being on the `stores` argument, not on the `getStates` name.
It could also be `getStates<stores>()` like we are doing with `getDispatch<models>()`

But this would not be enough.

> We would still need a way to get the models from the store. E.g. `store1.models`

I believe currently there is no way to get the `models` from a `store`. So you would need to add a way to do it. Only after that we would be able to add typings to your new `getState`.

brunolemos:
Closing as this PR was not made for multiple stores

PR #438, opened 2018-06-19
[loading] don't always resolve result

nihgwu:
right now the `loading` plugin always resolve result even if we reject in the effect, it resolves with `undefined`


ShMcK:
Excellent. Thank you for also including clear tests. :)

nihgwu:
@ShMcK mind to release a new version? this bug is annoying when there are errors in effects, it just swallows them

ShMcK:
Should be released now as v1.1.0

nihgwu:
thanks, upgraded

PR #441, opened 2018-06-20
Actions plugin to replace effects and selectors

bosung90:
I have been using rematch for the past 2 weeks.
This is the best redux library ever,
but I ran into a couple of annoying problems: 
1. calling effects plugin through dispatch.modelName.effectName not resolving synchronously, but rather dispatches an action that hopefully gets resolved by middleware. Also calling dispatch.modelName.effectName does not return what my function in effects returns and always wrap it in a Promise.
2. using selectors plugin is broken in 1.x.
3. Using selector plugin cannot reference your own or other selectors easily.

I created actions plugin that not only combines both effects and selectors plugin but also solves my problems stated above.

Actions is like combining mapStateToProps, mapDispatchToProps, and redux-thunk all at once.

To use actions plugin

### 1. add it as a plugin
```js
// src/store.js
import * as models from 'src/models'
import actionsPlugin from '@rematch/actions'
export const store = init({
  models,
  plugins: [actionsPlugin],
})
export const { dispatch, getState } = store
```

### 2. define actions inside your model
```js
// src/models/index.js

export const cart = {
 state: {},
 actions: ({dispatch, getState})=> {
    
 }
}

```

### 3. actions can call your own reducers, other reducers, and other actions zero, one, or many times synchronously or asynchronously.

```js
// src/models/index.js

export const cart = {
  state: 0,
  reducer: {
    increment: state => state + 1,
  },
  actions: ({ dispatch, getState }) => ({
    incrementMutiple(count) {
      for (let i = 0; i < count; i++) {
        dispatch.cart.increment()
      }
      // incrementMultiple will return synchronously with this string, no more promise.
      return `incremented ${count} times!!` 
    },
    async incrementAsync() {
      // wait 1 second before calling cart.increment
      await new Promise(r => setTimeout(r, 1000))
      dispatch.cart.increment()
      // incrementAsync will return promise which resolves to 'Succeed' 1 second later
      return 'Succeed!!' 
    },
    callOtherModelReducerOrActions() {
      dispatch.otherModel.reducerName()
      dispatch.otherModel.actionName()
    },
    // Object instead of a function? Actions plugin will dispatch this obj for you.
    simpleSingleDispatch: { type: 'cart/increment' }, 
    getCartState() {
       // Note that getState returns entire redux state.
      return getState().cart 
    },
    getCartStateMultiple(multiplier) {
      // You can call other actions that instantly returns some state
      return dispatch.cart.getCartState() * multiplier 
    },
  }),
}

```

### Here is a more complete example.

```js
// src/models/index.js
import { get, omit } from 'lodash'

export const cart = {
  state: {
    products: {},
  },
  reducers: {
    addProduct: (state, { id, price, name }) => ({
      ...state,
      products: {
        ...state.products,
        [id]: {
          id,
          count: get(state.products[id], 'count', 0) + 1,
          price,
          name,
        },
      },
    }),
    subtractProduct: (state, { id }) => ({
      ...state,
      products:
        state.products[id].count === 1
          ? omit(state.products, id)
          : {
              ...state.products,
              [id]: {
                ...state.products[id],
                count: state.products[id].count - 1,
              },
            },
    }),
  },
  actions: ({ dispatch, getState }) => ({
    getProducts() {
      return getState().cart.products
    },
    getSubTotal() {
      const products = dispatch.cart.getProducts()
      return Object.keys(products).reduce((prev, id) => {
        return prev + products[id].price * products[id].count
      }, 0)
    },
    getTax() {
      return dispatch.cart.getSubTotal() * 0.12
    },
    getTotal() {
      return dispatch.cart.getSubTotal() + dispatch.cart.getTax()
    },
  }),
}

```

### DIFF
--- a/plugins/actions/index.js
+++ b/plugins/actions/index.js
@@ -65,8 +65,8 @@ export default {
         )
 
         // Bind exposed dispatch[modelName][actionName] to call actions[actionName]
-        this.dispatch[model.name][actionName] = payload =>
-          actions[actionName].call(this.dispatch[model.name], payload)
+        this.dispatch[model.name][actionName] = (...params) =>
+          actions[actionName].apply(this.dispatch[model.name], params)
       } else if (typeof actions[actionName] === 'object') {
         // Bind exposed actions[modelName/actionName] to dispatch(actions[actionName])
         this.actions[`${model.name}/${actionName}`] = () =>
--- a/plugins/actions/index.js
+++ b/plugins/actions/index.js
@@ -1,0 +1,95 @@
+export default {
+  exposed: {
+    // access exposed actions obj using this.actions
+    actions: {},
+    storeDispatch() {
+      console.warn('Warning: store not yet loaded')
+    },
+    storeGetState() {
+      console.warn('Warning: store not yet loaded')
+    },
+    // Because we do not have access to store until the store is initialized,
+    // Create a reference point for dispatch and getState that onModel can call.
+    dispatch(action) {
+      return this.storeDispatch(action)
+    },
+    getState() {
+      return this.storeGetState()
+    },
+  },
+  onStoreCreated(store) {
+    this.storeDispatch = store.dispatch
+    this.storeGetState = store.getState
+  },
+  onModel(model) {
+    if (!model.actions) {
+      return
+    }
+    if (
+      typeof model.actions !== 'function' &&
+      typeof model.actions !== 'object'
+    ) {
+      throw new Error('actions must be a function or an object')
+    }
+
+    // model.actions can be a function that takes dispatch and getState as callback params
+    // or simply just an object (if you don't need to use dispatch and getState)
+    const actions =
+      typeof model.actions === 'function'
+        ? model.actions({
+            dispatch: this.dispatch,
+            getState: this.getState,
+          })
+        : model.actions
+
+    for (const actionName of Object.keys(actions)) {
+      this.validate([
+        [
+          !!actionName.match(/\//),
+          `Invalid action name (${model.name}/${actionName})`,
+        ],
+        [
+          typeof actions[actionName] !== 'function' &&
+            typeof actions[actionName] !== 'object',
+          `Invalid action (${
+            model.name
+          }/${actionName}). Must be a function or an object`,
+        ],
+      ])
+
+      if (typeof actions[actionName] === 'function') {
+        // Bind exposed actions[modelName/actionName] to call actions[actionName]
+        // actions[modelName/actionName] will be called from middleware
+        this.actions[`${model.name}/${actionName}`] = actions[actionName].bind(
+          this.dispatch[model.name]
+        )
+
+        // Bind exposed dispatch[modelName][actionName] to call actions[actionName]
+        this.dispatch[model.name][actionName] = payload =>
+          actions[actionName].call(this.dispatch[model.name], payload)
+      } else if (typeof actions[actionName] === 'object') {
+        // Bind exposed actions[modelName/actionName] to dispatch(actions[actionName])
+        this.actions[`${model.name}/${actionName}`] = () =>
+          this.dispatch(actions[actionName])
+
+        // Bind exposed dispatch[modelName][actionName] to dispatch(actions[actionName])
+        this.dispatch[model.name][actionName] = () =>
+          this.dispatch(actions[actionName])
+      }
+    }
+  },
+  middleware(store) {
+    return next => action => {
+      if (action.type in this.actions) {
+        next(action)
+        return this.actions[action.type](
+          action.payload,
+          store.getState(),
+          action.meta
+        )
+      } else {
+        return next(action)
+      }
+    }
+  },
+}


d3dc:
Whats stopping me from writing something real obtuse like:
```js
  wonkTheTotal(state) {
    const id = Object.keys(getState().cart)[0]
    await dispatch.cart.subtractProduct({ id })
    return state.cart.products
  },
```

State exists at 3 different points in time.

---

I'm also not sure how I safely consume an action in React. How do I know if I'm consuming it wrong or if it supposed to change something?

```js
connect(undefined, dispatch => ({
  total: dispatch.cart.wonkTheTotal
}))(({ total }) =>
  <h1>Current total: ${total()}</h1>
)
```

---

I think ultimately, I'd want to pose the question to you: **"Why do you need redux in your project?"**

bosung90:
@d3dc 
I think you have a really good point, but I think there is a big benefit over that slight trade-off.

I have been using redux for over 2 years, and I usually never confuse my selector to my other mapDispatch functions because I name all my selectors to start with 'get'.

I think the fact that you put it actions does not guarantee it to be side effect free, but my example shows that it can be used as selectors. If you are working with large number of people and require some sort of proof that it does not have side effects, I think existing selectors plugin is a great choice (with some slight improvements).

bosung90:
@d3dc I just checked out your PR for improving existing selectors. I haven't tried it out yet but it looks a lot better!! Now I am starting to see the benefits of selectors library 👍

ShMcK:
Close in favor of #428.

bosung90:
@ShMcK This plugin is not a replacement for select, I'm not sure why you closed it. I'm open to suggestions if you want me to rename this to something else or provide you examples on how I use it in my project.

ShMcK:
Sorry, my confusion. It seems to be:
1) similar to effects, but promises return a value
2) combines dispatch & select into one

I think the second feature may be quite inefficient. I may be wrong, but I believe you're essentially running every selector through a large set of action pattern matchers. Also bear in mind that store subscriptions such as `connect` listen to every dispatch, and may trigger additional re-renders.

bosung90:
I use this plugin exclusively and I don't really see a problem. Being able to dispatch based on a state is essentially what redux-thunk does, and being able to select part of a state without requiring a connect is a big win. For instance, you may want to fetch some state when someone taps on a button. You still need to use some sort of connect to render based on the state, but isn't that exactly same as existing select plugin?
I think rematch is awesome because it saves so much time and not have to write a lot of boilerplate code, being able to create plugin really truly maximizes the re-usability across multiple different projects. I believe this plugin really saved me a lot of time and boilerplate code by just giving maximum flexibility of 
1. synchronous call, returns whatever function returns (same as redux's mapDispatchToProps), but still able to do asynchronous.
2. access to dispatch and getState (redux-thunk)
3. able to call actions without using connect (dispatch.cart.actionName)

d3dc:
@bosung90 What you've done here is pretty cool. You've essentially built your store slices as evented or almost observable services. Whats interesting is that its really the facade pattern over a global state.

You may or may not recognize "observable services" as an angular thing. There you request a service by its token (name), and get a singleton that has observable members:
```js
export class ProductService {
  products = Subject<object>
 
  constructor(private messageService: MessageService) { }  

  addProduct (product: Product) {
    this.products.next(
      products.map(state => ({ ...state, product }))
    )
  }

  getSales(): Observable<Product[]> {
    this.messageService.add('ProductService: fetched products');
    return of(api.fetchSales());
  }
}
```

```js
export class Cart extends Component {
  constructor(private productService: ProductService) { }
  ...
}
```

---

That leaves me with a few comments, only the first really negative.

---

**Does your PR achieve the same goal as why we use [Flux](https://facebook.github.io/flux/docs/in-depth-overview.html#content)?**

> Being able to dispatch based on a state is essentially what redux-thunk does, and being able to select part of a state without requiring a connect is a big win.

If you look for it in the docs, you might notice an undertone of "thunk is bad". This is because of that `getState()` escape hatch. Data flow _is not forced_ to be unidirectional. It becomes a footgun. I commented my first post to illustrate this.

---

**Is there an advantage to the facade pattern?**
If we never see `state`, and it falls in the woods, does it make a sound?

Looking at the angular example, state is actually stored on the class instances. 

This is achievable because there's dependency injection wiring going on. Angular has some pretty heavy wiring here so that the services can be used in constructors.

---

I hate to bring a different PR into this, but obviously a lot of this stuff has been on my mind while writing the select plugin. You'll notice I call the reducers and selectors the "getters" and "setters" of the store slice, respectively. I'm also going for a service facade over the state, but still following Flux.

a utility I've written internally looks like:
```
withModels('cart', 'user', '$views')(MyComponent)
```

I think this helps force everyone to write Flux when its the **chosen** architecture pattern.

Would your mental model translate to something like that? Could you build your models with their setters and getter separated in the configuration, but then injected as if they weren't?  IMO, this would be a great plugin to build-out - **on top of** other plugins that do the translating to redux.

`store.service` anyone?

---

And just to throw another project idea out there. An entirely different pain point I touched on is the DI factories. 

Could models be written in such a way as to participate in `withModels` or have a top-level DI? I've played around with this, and you could do it using ES6 Proxy objects that eventually become the "models".

ShMcK:
closing due to inactivity

PR #446, opened 2018-06-23
remove global dispatch/getState

### LINKED ISSUE #439: Remove globals: `getState`, `dispatch`?
There are a few issues with `getState` in the current v1.

1. The name implies it should be a single store's state. It may be better to call it `getStates`

2. How to handle TS typings if it returns multiple stores states. Typings are currently created from models for a store, rather than for collection of stores.

ShMcK:
closes #439. 

Let's face it. A global dispatch & global getState were a bad idea. They produce problems such as:

- memory leaks
- inability to use multiple stores
- encourage large monolithic apps
- impossible to add proper typings

Globals can be easily avoided though:

```js
import { init } from '@rematch/core'

export const { getState, dispatch } = init()
```

Just import your `getState` & `dispatch` from elsewhere.

### Todos
- [x] remove global dispatch/getState
- [x] remove global typings
- [x] add warnings that explain what to do in place of `getState` and `dispatch` calls
- [x] make breaking changes explicitly clear in the readme
- [x] remove references to global dispatch/getState from docs
- [x] remove references to global dispatch/getState from examples
- [x] respond to people asking "WHY DID YOU BREAK MY APP? I love globals!"


krainboltgreene:
Fantastic work!!!

PR #447, opened 2018-06-23
Upgrade react-navigation plugin to react-navigation v2

d3dc:
This is a fork of @bosung90's work to make `react-navigation` v2 work.

I changed the api to take a `Navigator` instead of "Routes" and return a `ConnectedNavigator`, this seems to align better with the examples for react-navigation. It puts the two create calls next to each other.

I think the ideal way to add functionality is through a model, as it has the least surface area.

This also a includes selectors for the selectors in #428 on the model. I believe this is the right way to do this, as the config won't be used otherwise, and it abstracts the inner workings entirely.


bosung90:
Looks really good, as long as it works I am happy haha! Thank you for all the work. I feel really bad for not finishing my work but now you made it even better.

d3dc:
@bosung90 I'm not sure I've ever written the part where I would have to write `<Routes />` before. Am I trampling some tradition where the config and Navigator are combined? (Just in the README and example)

d3dc:
@ShMcK @bosung90  Turns out there's no tests here...

Uh. The store model will replace the redux reducer. I see the issue here now...

It could just use a different key.

I wonder if it would be better to check for `this.reducers[model.name]` and use it instead of just returning `state` in `createModelReducer`

ShMcK:
I agree, #450 is an elegant solution. I'll leave some notes there.

d3dc:
While #450 balloons, I reverted this back to just replacing the dispatcher at `storeKey`. If `store.select` exists, it will also replace the selectors for `storeKey`

ShMcK:
Would you say this PR is ready to merge?

d3dc:
@ShMcK I was having trouble getting the example to install.

I tried writing a test case, but jest just wasn't transforming `react-navigation-redux-helpers` no matter where I put `transformIgnorePatterns`

ShMcK:
I've had issues before with the React-Navigation plugin and "peerDependencies" & `npm link`, due to the way npm link works in NPM 5. If you think this is related, let me know and I'll try to help out.

d3dc:
I think the transform isn't happening because its `jest-typescript` not `jest-babel` and `react-navigation-redux-helpers` needs babel.

d3dc:
@ShMcK I'm not sure how to get the example to run without a published version. I tried a file path and a github path, and both did not run.

ShMcK:
@d3dc this is likely due to peerDependencies and how `npm link` fails to find them if the module path is outside the directory. 

I assume the package works if it's inside of the project?

d3dc:
@ShMcK If I move the source into the example and convert it to plain javascript, the packager just fails with:
> Failed building JavaScript bundle
and no other output

I'll have to go through the example and see if theres some compiler options or something.

d3dc:
So it looks like Effects aren't bound to dispatch here or in the recipe I added. Did I miss this, or am I missing something now?

It should be using bindActionCreators, no?
```
effects: dispatch => bindActionCreators({ ... }, dispatch)
```

ShMcK:
Any progress on this PR? I know its been a while.

bosung90:
react-navigation no longer support redux integration. From them:
```
Redux integration
Warning: in the next major version of React Navigation, to be released in Fall 2018, we will no longer provide any information about how to integrate with Redux and it may cease to work.
```

ShMcK:
I support that decision. In my own work, I've moved to using `this.props.navigation` directly with performance increases.

Issue #414, opened 2018-06-12
Listen for actions dispatched from redux reducers?


alozytskyi:
While migrating to Rematch, there is still significant chunk of the state which is managed by the plain reducers and actions, however Rematch doesn't seem to react to them

```javascript
reducers: {
  "REDUX_ACTION": function(state, params) {
    // react to action
  }
}
```
neither as
```javascript
reducers: {
  "REDUX_ACTION"(state, params) {
    // react to action
  }
}
```
However, to ensure seamless transition it would be useful to listen to the reducers that are yet pending  to be converted into models. It might be that plain names won't work here, but what if it'd be something like `__redux/REDUX_ACTION`, a special case?

alozytskyi:
This seems to pass the test:

```patch
Index: test/listener.test.js
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- test/listener.test.js	(revision 44713f84827d8d9ddbea4fd47e1ec4e510aaafe4)
+++ test/listener.test.js	(date 1528797161000)
@@ -24,4 +24,21 @@
 
     expect(store.getState()).toEqual({ count1: 1, count2: 1 })
   })
+
+  test('should trigger state changes on redux actions', () => {
+    const count = {
+      state: 0,
+      reducers: {
+        '__redux/INCREMENT': (state, payload) => state + payload
+      }
+    }
+
+    const store = init({
+      models: { count }
+    })
+
+    store.dispatch({type: 'INCREMENT', payload: 1});
+
+    expect(store.getState()).toEqual({ count: 1})
+  })
 })
\ No newline at end of file
Index: src/redux.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/redux.ts	(revision 44713f84827d8d9ddbea4fd47e1ec4e510aaafe4)
+++ src/redux.ts	(date 1528797295000)
@@ -36,9 +36,14 @@
 	this.createModelReducer = (model: R.Model) => {
 		const modelReducers = {}
 		for (const modelReducer of Object.keys(model.reducers || {})) {
-			const action = isListener(modelReducer)
+			let action = isListener(modelReducer)
 				? modelReducer
 				: `${model.name}/${modelReducer}`
+
+			if (/^__redux\//.test(action)) {
+				action = action.replace(/__redux\//, '')
+			}
+
[...]

Issue #409, opened 2018-06-11
Use with connected-react-router?


linonetwo:
`react-router-redux` is deprecated https://github.com/supasate/connected-react-router/issues/74
It's preferred to use `connected-react-router`.

But to set up `connected-react-router` we need to pass `rootReducer` to it. https://github.com/supasate/connected-react-router#step-1

So to work with `connected-react-router`, I need to write a plugin?

ShMcK:
I don't think you ever need to write a plugin. Plugins are just a way to make a configuration reusable.

These `redux.rootReducer` docs may help: https://github.com/rematch/rematch/blob/master/docs/reduxApi.md#rootreducers

linonetwo:
But I need to pass `rootReducer` to it
```js
redux: {
        middlewares: [routerMiddleware],
        rootReducers: {
          router: connectRouter(history)(rootReducers)
        },
      },
```
But where can I get this `rootReducers`?
Maybe it's `connected-react-router`'s fault since it's requiring itself to be the root reducer  https://github.com/supasate/connected-react-router/blob/master/src/reducer.js#L37

d3dc:
@linonetwo You can already listen to rootReducers ;)

```js
import { combineReducers } from 'redux'
import { connectRouter, routerMiddleware } from 'connected-react-router'
import { createBrowserHistory } from 'history'
const history = createBrowserHistory()

export default enhanceWithConnectRouter = (config) => ({
  ...config,
  combineReducers: (rootReducers) => connectRouter(history)(combineReducers(rootReducers)),
  middlewares: [...config.middlewares, routerMiddleware(history)]
})
```

linonetwo:
But as https://github.com/rematch/rematch/issues/376#issuecomment-394731320 said, if I want to listen on router change, then loads data according to the new route, I need
```js
{
  effects: {
  ['router/PUSH'](state) {
    ...
    return state;
  },
  },
}
```
It will cause `Invalid effect name` at this time.

YanYuanFE:
```
import { init } from '@rematch/core';
import { connectRouter, routerMiddleware } from 'connected-react-router';
import createLoadingPlugin from '@rematch/loading';
import createHistory from 'history/createHashHistory';

const options = {};
const loading = createLoadingPlugin(options);

export const history = createHistory();

const reducers = { router: connectRouter(history) };

/**
 * 基于 Rematch 实现的 Redux 最佳实践
 *
 * More: https://rematch.gitbook.io/handbook/
[...]

PR #450, opened 2018-06-23
Add `reduxReducer` config to `Model`

d3dc:
Writing #447, I wanted to use a model to expose the effects and selectors of an existing library instead of modifying the store.

To provide a library reducer, we have to put it in `redux.reducers` - which will get replaced by a model of the same name.

I also wrote a little plugin for #414, which just lifted the redux actions to the rematch format, _as if they came from a model_. Writing a model implicitly lifts the libraries actions.

The only downside is that the library actions then have to be `effects`, which are async.


ShMcK:
Good idea. I've had use cases for this as well when using libraries.

We will also have to update the docs.

It seems there's a formatting issue in redux.test.js. Do you have prettier enabled?

ShMcK:
Another option I had considered was:

If `reducers` is a function, treat it as a redux reducer.

d3dc:
I do have prettier on. I saw the .prettier file and just let it go. That might be my mistake.

Reducers as a function was my first (second?) thought in the other PR. After thinking about it, it seemed awkward to have different functionality depending on type. 

I would think of it as wrapping or adapting a redux reducer. You port a redux slice to rematch by writing an adaptor model. And I would wonder if being able to add additional state mutations on top is actually a bad thing.


... It could also be a functional way to derive state. `model.state` is a function?

d3dc:
I think a descriptive option name would be the only change needed.
`baseReducer`, `legacyReducer`, `oldStyleReducer`

ShMcK:
Ah, so you imagine a `model.reduxReducer` and `model.reducers` could be used together?

I'd love to see an example of how it would work.

d3dc:
One use case for a legacy reducer is to gradually upgrade. Half your reducer can exist in a function, and the other half in rematch.

ShMcK:
A) model.reducers (object)
B) model.reduxReducer (function)

Sounds like a good idea. Currently the functionality in this PR is A || B:

```js
return modelReduxReducer ? modelReduxReducer(state, action) : state
```

The next state can be the result of A & B composed together, but that leaves the question:

- which is calculated first? A or B? How can this be made clear?

d3dc:
The reason I mentioned the `state` function above is because that direction seems the most appropriate - The model reducers run first and the library is still responsible for the final representation of the data.

I did A || B here, because that is how almost all reducers are written. The model reducers in this library will return if one matches; most switch-case reducers return.

If you think about it as a type system where reducers are the "setter methods", action A on the model should shadow action A on the base, no?

d3dc:
@ShMcK You're probably right here. `baseReducer` should just be a smaller `rootReducer`. 

I went ahead and added a recipe and documented `baseReducer` anything else needed here?

d3dc:
How do I listen to `LOCATION_CHANGE` from `react-router-redux`? `router/push` exists, but what about after route change? What if I want to catch `goBack` etc.?

ShMcK:
Had some time to play around with the baseReducer this morning. Looks good. Hopefully we can start closing your PRs soon.

ShMcK:
I'm not sure about using "react-router-redux". Currently I've been using [reach-router](https://reach.tech/router), and not binding the router and redux together in my own projects.

It seems [react-router-redux](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux) is no longer maintained and [connected-react-router](https://github.com/supasate/connected-react-router) is now the recommended path. It may worth posting an issue.

ShMcK:
OK to merge?

d3dc:
`connected-react-router` has #409 and does not lend itself to a simple example. It takes over the state slicing for some reason, and needs to have access to the entire store state.

Other than having a deprecated library in the example this should all be good.

ShMcK:
published as 1.0.0-beta.5

PR #455, opened 2018-06-25
fix relative import of core typings in loading plugin

andrewheekin:
relative imports are currently breaking due to an incorrect path


ShMcK:
Thanks :)

PR #461, opened 2018-07-02
Persist Plugin: Optional callback when creating persistor.

guidotajan:
I work with '_react-native-navigation_' and I need to know when the persistor has done its job without using the PersistorGate.
The callback added here does the trick.

### DIFF
--- a/plugins/persist/src/index.ts
+++ b/plugins/persist/src/index.ts
@@ -8,7 +8,7 @@ let persistor
 export const getPersistor = () => persistor
 
 // rematch plugin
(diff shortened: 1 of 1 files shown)


ShMcK:
That's great. Maybe we should flush the whole `persistStore` config out as well.

```js
const persistPlugin = (config = {}, pesistStoreConfig, callback): Plugin => {
```

```js
persistor = persistStore(store, persistStoreConfig, callback)
```

What are your thoughts?

guidotajan:
Sure! Great idea. I just committed the changes.

Issue #464, opened 2018-07-04
effect's rootState vs redux-thunk's getState 


glimow:
When using redux-thunk, your async actions have a `getState()` parameter to know about the current state. In rematch, `rootState` is not a function but an object, thus it can be outdated when using several reducers in the same effect.
```javascript
effects: (dispatch)=>({
    async fetch(payload, rootState){
          await dispatch.model.someReducer(payload);
          //rootState is still the old state here 
    },
})
```
 Since `rematch 1.0.0` it is not possible to import global `getState()` and importing store's `getState()` would make a circular dependency as we are in a model.
Is dispatching several actions in one effect a bad practice ? Or is there a workaround ? Why not passing both `dispatch` and `getState` to the effects function, and remove `rootState` ?

ShMcK:
You raise a good point. If there is a possibility of getting the wrong state, `rootState` may be a bad idea.

Internally, rootState is created by a `getState()` call and passed in as a param when the effect is processed. Have you seen it have the wrong value? If it can be demonstrated, we definitely need to make a change.

---

## Alternatives

I can think of a few alternatives to removing getState that you might find useful.

### Import `getState`

It's possible to import `getState` within an effect. 

```js
import { getState } from './store'

export default {
  effects: (dispatch)=>({
    async fetch(payload){
         console.log(getState())
          await dispatch.model.someReducer(payload);
          //rootState is still the old state here 
    },
  })
}
```

### PR to effects function

It may even make more sense to add `getState` as a second param.

```js
 effects: (dispatch, getState)=>({
```

I'd be happy to accept any PR that adds `store.getState` as the second option.

jackwilsdon:
I like the idea of passing `getState` as a second argument - it more closely follows how `redux-thunk` works.

glimow:
@ShMcK I just ran on several cases where rootState had the wrong value when migrating from redux-thunk code to rematch. I guess that's because in redux, state is treated as immutable and is replaced at each dispatch.

I tried your first solution, but that made a circular dependency, model=>store=>model and store resolved to `undefined`. At least I guess because I couldn't find any other explanation for it.

Second solution definitely looks nice and smooth. Will try to do a PR for it.

slashwhatever:
How did this issue die? Never implemented?

pjamrozowicz:
It was never implemented but it could be a nice addition to the v2 PR if someone wants to join me in improving rematch:
https://github.com/rematch/rematch/pull/720

[...]

PR #465, opened 2018-07-04
Add getState to function effects

glimow:
following discussion about https://github.com/rematch/rematch/issues/464,
this PR adds `getState` to effects-function parameters as that avoid deprecated state and looks more like `redux-thunk`'s syntax
```javascript
effects: (dispatch, getState) => ({
		async asyncAddOneArrow() {
			console.log(getState) // { example: 1 }
			await dispatch.example.addOne()
			console.log(getState) // { example: 2 }
		},
	}),
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -124,6 +124,7 @@ Reducers may also listen to actions from other models by listing the 'model name
 #### effects
 
 `effects: { [string]: (payload, rootState) }`
+or `effects: (dispatch, getState) => ({ [string]: (payload, rootState) })`
 
 An object of functions that can handle the world outside of the model.
 
@@ -148,6 +149,23 @@ Effects provide a simple way of handling async actions when used with `async/awa
       const data = await response.json()
       // pass the result to a local reducer
       dispatch.example.update(data)
+    }
+  }
+}
+```
+
+Effects can be defined as a function that recieve `dispatch` and `getState` as arguments.
+
+```js
+{
+  effects:(dispatch, getState) => {
+    async loadData(payload) {
+      // wait for data to load
+      const response = await fetch('http://example.com/data')
+      const data = await response.json()
+      // pass the result to a local reducer
+      dispatch.example.update(data)
+      const updatedRootState = getState()
     }
   }
 }
--- a/src/plugins/dispatch.ts
+++ b/src/plugins/dispatch.ts
@@ -26,6 +26,10 @@ const dispatchPlugin: R.Plugin = {
 			return this.storeDispatch(action)
 		},
 
+
+		getState(){
+			return this.storeGetState()
+		},
 		/**
 		 * createDispatcher
 		 *
--- a/src/plugins/effects.ts
+++ b/src/plugins/effects.ts
@@ -20,7 +20,7 @@ const effectsPlugin: R.Plugin = {
 
 		const effects =
 			typeof model.effects === 'function'
-				? model.effects(this.dispatch)
+				? model.effects(this.dispatch, this.getState)
 				: model.effects
 
 		for (const effectName of Object.keys(effects)) {
--- a/src/typings.d.ts
+++ b/src/typings.d.ts
@@ -159,7 +159,7 @@ export interface ModelConfig<S = any, SS = S> {
   state: S,
   baseReducer?: (state: SS, action: Action) => SS,
   reducers?: ModelReducers<S>,
-  effects?: ModelEffects<any> | ((dispatch: RematchDispatch, getState():any) => ModelEffects<any>),
+  effects?: ModelEffects<any> | ((dispatch: RematchDispatch, getState:()=>RematchRootState<any>) => ModelEffects<any>),
   selectors?: {
     [key: string]: (state: SS, ...args: any[]) => any,
   },
--- a/src/typings.d.ts
+++ b/src/typings.d.ts
@@ -159,7 +159,7 @@ export interface ModelConfig<S = any, SS = S> {
   state: S,
   baseReducer?: (state: SS, action: Action) => SS,
   reducers?: ModelReducers<S>,
-  effects?: ModelEffects<any> | ((dispatch: RematchDispatch) => ModelEffects<any>),
+  effects?: ModelEffects<any> | ((dispatch: RematchDispatch, getState():any) => ModelEffects<any>),
   selectors?: {
     [key: string]: (state: SS, ...args: any[]) => any,
   },
--- a/test/effects.test.js
+++ b/test/effects.test.js
@@ -385,5 +385,33 @@ describe('effects:', () => {
 				example: 1,
 			})
 		})
+
+		it('should pass getState in as a function', async () => {
+			const example = {
+				state: 0,
+				reducers: {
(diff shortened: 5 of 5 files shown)


ShMcK:
Seems to be a TS related issue. Will also have to update:

- [ ] typings
- [ ] docs

d3dc:
@glimow I kind of disagree with having a `getState` function. I think your effects shouldn't be miniature redux loops, but just grouping and describing the flow of actions.

If you need to depend on the result of an action, you shouldn't listen to the store, but rather the effect. This is very easy to do since effects are promises. `getState` encourages using it as an escape hatch from the effects pattern.

I don't have any weight in this, but a thought I've been having about how this should work, is that effects should just receive all the store's plugin properties or similar.

You could easily introduce a `getState` plugin for this functionality.

```js
const model = {
  effects: (dispatch, getState, select) => ({
    ...
  })
}

const getStatePlugin = () => ({
  onStoreCreated(store) {
    return {
      getState: store.getState // Set it to the same value but now knows about it?
    }
  }
})

const store = {
  models: { model }
  plugins: [getStatePlugin(), selectPlugin()]
}
```

glimow:
@d3dc have you an example of a situation where it could lead to problems ? `redux-thunk` does it this way and it works fine. I see no point in having effects to access dispatch and not getState: it means that you are blind to your own state updates inside them.
@ShMcK I'm on it

d3dc:
I think my real point is: Can this be a plugin? If I don't use `getState` because I'm a pedant, I'll still have to include it if I want to use something like `extraArguments` (which does seem useful).

ShMcK:
Let me summarize to see if I'm on track.

`effects: (dispatch, getState, extraArguments) => ({ ... })`

or 

`effects: (dispatch, { getState, select }) => ({ ... })`

If I had to choose, I think it may make more sense just to pass in the store:

`effects: (store) => ({ ... })`

`store.dispatch`, `store.getState` & `store.select` are all available.

This is, however, a breaking change. But that's why we keep things in beta :)

d3dc:
@ShMcK I'm a fan of explicit arguments because it makes it clear what should be happening.

A function that takes store does something with store. A function that takes dispatch, will dispatch actions. I'm fudging a little bit with the extra args part, it would make sense if I was handed skeleton code with the extractor, but so would the store.

I also think it could be bad if someone say, `subscribe`d in an effect.

glimow:
I see no real issue in having getState in effects. Whether you use it or not, you will always have state history to debug. Plus accessing the store in effects was possible before global getState and dispatch removal, so I guess there could be a bunch of rematch users that already do it that way.

d3dc:
@glimow would it be possible to get some examples of effects needing `getState`? 

I'll try not to pick em to shreds 😅

d3dc:
```js
effects: (dispatch, {getState, ...extraArgs})=>{}
```

Feels right to me. Is the debate really just whether the dispatch plugin has extraArgs?

To complement my last comment, it seems like the pattern from the original issue can be avoided by our old friend, the coordinating effect. 

```js
async storeSecurely(document) {
  await dispatch.credentials.refresh()
  return this.store(document)
}

store(document, state) {
  // state has fresh credentials already
  return api.documents.put(document, { key: state.credentials.s3 })
}
```

glimow:
@d3dc That's the workaround I used for now. But It's not really readable and it forces you to create unnecessary functions.
```
async storeSecurely(document) {
  await dispatch.credentials.refresh()
  const state = getState()
  await const { data } =  api.documents.put(document, { key: state.credentials.s3 })
  return this.documentSynced({ document, data })
}
```
seems easier to read.
This is a good use case example btw.

d3dc:
> That's the workaround I used for now.

Thats the proper feature to use! Its an actual pattern in the CQRS world. The framework does things this way for the change log reason. There's one action per dispatch call **and** one `getState`.


Consider the change log from your example:
```console
action @ 19:50:14.613 documents/storeAndSync (in 10.70 ms)
action @ 19:50:14.644 credentials/refresh (in 0.20 ms)
Uncaught Error: credentials not provided
    at vendors~app.b111668a9e5b59522cd9.js?b111668a9e5b59522cd9:37
```

Ah, the credential refresh must be failing! Wait...


If you instead order your action correctly (the best practice the framework is forcing you to do)
```console
action @ 19:50:14.613 documents/storeAndSync (in 10.70 ms)
action @ 19:50:14.644 credentials/refresh (in 0.20 ms)
action @ 19:50:14.674 documents/storeSecurely (in 0.12 ms)
Uncaught Error: credentials not provided
    at vendors~app.b111668a9e5b59522cd9.js?b111668a9e5b59522cd9:37

```

**Now I can clearly see that refresh succeeded and we started doing something else before it crashed.**

---

I'm actually starting to think my example is crappy. I cheated myself by choosing very un-descriptive names the first time - using better names makes it clear that you're naming the branches of your execution. Such a simple example also doesn't show error handling or branches very well. Try putting 3 `try-catches` into one thunk and figure out what threw the error. I really think it takes the cognitive load off to just handle errors of one type in each function with no internal state.

glimow:
Well, the framework does not do that only for logging purposes but also to avoid side effects. Here, getState or not, main state is still changed by pure functions aka reducers. I find the idea of making the code less readable only for logging purposes very strange. I'm pretty sure there are workarounds to keep good logging with the first method. Also, if we do like @ShMcK's suggestion, you can keep the old syntax with `effects: ({dispatch}, externals)`

d3dc:
> I find the idea of making the code less readable only for logging purpos very strange.

First, in what way is having a named function less readable? Its less code, and it names the steps of execution.

Second, as you said, its for side-effects. Logging is just for us developers, the rest of the app still follows the store by it's actions. I'm doing a bad with these toy examples, but see my previous point about error handling. You don't want to mix errors from different actions, but you do want them to bubble up.

> I'm pretty sure there are workarounds to keep good logging with the first method.

**"Invent Workarounds"** 

The way to do this would be to dispatch dummy actions. However, as far as I can tell, _in this framework_ actions == dispatchers == functions. If you need a workaround, you are fighting it...

I've been trying to think an analogy to React's `setState`.  Models are almost the components of the store, and `dispatch` is your `setState`. In React, you don't have a `getState` equivalent, you listen to `setState`.

> Also, if we do like @ShMcK's suggestion, you can keep the old syntax with effects: ({dispatch}, externals)

I absolutely hate this syntax. Its not obvious and its dangerous. If you `pick` store by default, you might as well have positional arguments

ShMcK:
I'm starting to wonder if this is really necessary, as `getState` can always be imported when necessary.

```js
import { getState } from './store'
```

I'm worried about adding complexity to the API if it's not totally necessary.

d3dc:
@ShMck That would create a circular dependency between the model and store, no?

ShMcK:
I suppose this is what happens when you leave comments at the end of the day :)

You're right.

ShMcK:
Thanks everyone for your input on this. 

Unfortunately, will close to keep the API as simple as possible. If you need to ensure you are using the most up to date store state, you can alternately use effects that call other effects.

gaoqiang19514:
hurry up, please！

Issue #221, opened 2018-02-22
Plugin: Strictly type check model state


ShMcK:
It would be great to have a plugin that checks reducer state and validates it against a set of types. This is feature that exists in MobX, but isn't available in Redux (without using TypeScript).

For example:

```js
import { init } from '@rematch/core'
import rematchStateTypingsPlugin from '@rematch/stateTypings'

const count = {
  state: {
    name: '',
    age: null
  },
  typings: {
    name: 'string',
    age: 'number|null'
  },
  reducers: { ... },
}

const store = init({
  models: { count },
  plugins: [rematchStateTypingsPlugin]
})
```

On each state change (in development mode), warnings could be triggered in the console if the next state is not properly validated against the state typings.

This is currently just an idea with a rough outline. I'd love to hear anyone else's thoughts, or whether or not this idea would be useful.

Ne0nx3r0:
I tried to do this using implied typings so you could maintain your actions/reducers and immediately see where things break. 

The issue I kept running into is without either explicitly defining your action types or a fairly unpalatable amount of boilerplate the implied type ends up getting lost and you lose that benefit.

TypeScript has generics and implied types, but I couldn't find a way to define an implied type so you can reference it explicitly without redefining the type.

An alternative for pure Redux is something like this: https://spin.atomicobject.com/2017/07/24/redux-action-pattern-typescript/

But it's quite verbose and falls into that boilerplate problem I mentioned.

ShMcK:
It would definitely be preferable to to validate against a reference to a TypeScript or Flow type.

If there is strong demand for this kind of plugin, I'd consider looking into it.

will-stone:
This reminds me of mobx-state-tree. Typings would certainly add to the power and allure of this package over straight Redux. I’d say go for it if you have the time available to develop it.

koss-lebedev:
Closed in #466

PR #466, opened 2018-07-05
Plugin for type-checking state

koss-lebedev:
Hey @ShMcK ,

I really like the idea you outlined in https://github.com/rematch/rematch/issues/221, so I started implementing it as a plugin. I wanted to submit PR early so that you could weigh in on things you don't like or disagree on.

First of all, I don't think it's possible to check state against TypeScript or Flow definitions since those are tools for static type checking and they won't be available at runtime. It would be super cool to be able to reuse TS/Flow definitions, but it simply won't work here.

So the next best thing I think we can do is to have our own type definitions that use syntax similar to that of Flow or TS. I know that `mobx-state-tree` uses custom API like this:

```js
const Todo = types.model({
    title: types.string
})
```

However, in our case, I don't think it's necessary, and simple string-based syntax definition will do just fine. So this is the example of how definitions can look like:

```js
typings: {
    name: 'string',
    age: 'number',
    height: 'number?',
    isDeveloper: 'boolean',
    birthDate: 'date|number',
    address: {
      country: 'string?',
      city: 'string',
    }
  }
```

It would support primitive types like `string` and `number`, union types like `string|number` which would be valid if the value is either `string` or `number`, and optional types like `string?` which is essentially equal to `string|undefined`.


ShMcK:
@koss-lebedev thanks for taking up the cause :)

Is there a schema validation library you were thinking of using for this? 

Validating types can be a can of worms once you dive into it, consider deeply nested objects, arrays with multiple types, immutable data structures, etc. Somebody's done the work for you.

koss-lebedev:
@ShMcK Thanks for your input! I actually agree with you that we should use ready solution for typechecking instead of reinventing it as I originally tried to do. One option I'm exploring now is reusing facebook's `prop-types` library. I see quite a few benefits: 

- It's been designed to typecheck objects _at runtime_, which is exactly what we need here
- it is battle tested
- a lot of developers are already using it, so that's a huge plus

There are some issues with `prop-types` API since it wasn't designed to work with anything else other than React  props, but I think there's a way around it. I'll let you know when I make it work

koss-lebedev:
@ShMcK 
OK, I think this is ready for review. I finished the plugin, added tests and documentation. Please let me know what you think :)

ShMcK:
This is really good. Great work.

I'm a bit worried about the package `ReactPropTypesSecret`, based on the constant names:

```
 __SECRET_DOM_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: ReactDOM,
  __SECRET_DOM_SERVER_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: ReactDOMServer,

ShMcK:
Publishing as `@rematch/typings` may be confusing, as it sounds like a typescript or flow package.

Any recommendations for an alternate name?

koss-lebedev:
That's a good point. I think we could use `typed-state` or `state-types` (kind of like `prop-types`), or maybe you have a better suggestion? Let me know and I'll submit a new PR with the updated names

ShMcK:
`typed-state` sounds clearer to me. Also if you could change the folder name as well :)

Thanks again!

Issue #440, opened 2018-06-20
How to use with redux-thunk?


krainboltgreene:
I'm setting up with:

```
import {init} from "@rematch/core"
import thunk from "redux-thunk"
import logger from "redux-logger"
import environment from "@internal/environment"
import * as models from "./models"
import sdk from "./sdk"

export default init({
  models,
  plugins: [],
  redux: {
    middlewares: [
      thunk.withExtraArgument({client: sdk}),
      logger,
    ],
  },
})
```

and my model:

```
export default {
  state: {},
  reducers: {
    updateFormInputValue (state, payload) {
      const {formSlug} = payload
      const {fieldSlug} = payload
      const {value} = payload

      return mergeDeepRight(
        state
      )(
        objectFrom(
          [formSlug, fieldSlug]
        )(
          {writtenAt: new Date(), value}
        )
      )
    }
  },
  effects: {
    submitRegistration (payload, root) {
      return function thunk (dispatch, {client}) {
		console.log(arguments)
	  }
    }
  }
}
```

However when `submitRegistration()` is dispatched my inner function doesn't get called?

d3dc:
Hey @krainboltgreene, `redux-thunk` is actually not needed for rematch. The new betas landed PR #399 and it hasn't made its way to the gitbook yet.

krainboltgreene:
Does this account for `withExtraArgument()`?

krainboltgreene:
Further this seems like a non-ideal interface? Now when I have one effect that needs dispatch, all effects now have access dispatch. It also means `effects` can't just be an object.

d3dc:
I'm not sure of the advantage of effects just being an object. Rematch will collect an object from factories, and the typing will extract the typings of that object.

Just anecdotal, but most of my dispatch heavy sagas occurred in the same models and before Rematch I wrote a whole lot more thunk boilerplate.

I think ExtraArguments might be a great idea to add if its not there already. At the worst, its a clone of the loading plugin that overloads args. _To beg the question_, what does `sdk` have to do with your application state?

krainboltgreene:
> I'm not sure of the advantage of effects just being an object.

For one it makes splitting up harder:

```
import * as effects from "./effects"

export default {
  state: {},
  reducers: {},
  effects,
}
```

vs 

```
import * as effects from "./effects"

export default {
  state: {},
  reducers: {},
  effects: (dispatch) => effects
}
```

The latter doesn't work.

[...]

PR #467, opened 2018-07-05
Add `extraArguments` property to plugins to expose extra arguments for model member factories

d3dc:
This PR adds the `extraArguments` property to plugins.

```js
const getStatePlugin = () => ({
  extraArguments: {
    getState () {
      return this.storeGetState()
    }
  }
})

const model = {
  effects: (dispatch, { getState }) => ({
    ...
  })
}

const store = init({
  models: { model },
  plugins: [getStatePlugin()]
})
```

After seeing #465, I'm hesitant to see the API change towards something that I think is a footgun to even have. There's a lot of people overthinking their routing already. #440 brought up a great point about a different feature from `thunks`, the `extraArguments`.

### DIFF
--- a/src/pluginFactory.ts
+++ b/src/pluginFactory.ts
@@ -14,7 +14,7 @@ export default (config: R.Config) => ({
 	 * bind validate to the store for easy access
 	 */
 	validate,
-
+	extraArguments: {},
 	/**
 	 * create plugin
 	 *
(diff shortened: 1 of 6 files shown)


d3dc:
I'm prefacing with "for action creators" because a plugin like my `select` couldn't be used before model initialization, but action creators aren't called until well into the application.

ShMcK:
I wonder if it might just be easier to pass `store` in as a single param. Keep everything in one place. I figure, if the best name for a param is `extraArguments`, it could be worth rethinking things.

It would still allow you to use `store.select`, as it should be bound to the store. See discussion in #465

d3dc:
@ShMcK `store` isn't available in the effects factory.

ShMcK:
Thanks @d3dc for all the work. I worry that this may add unnecessary complication to the API at this time. Will close as wontfix

d3dc:
I've been using an external DI container for this. It's more general, just as simple, and doesn't bloat an API.

```js
@inject('getState')
anEffect(state, payload, getState) { ... }
```

Using a proxy let's you wrap the parent instead and is all Rematch would be doing anyway (proxy fn). I don't doubt Typescript is even cleaner.

ShMcK:
Very interesting, love to see how your implementation works.

d3dc:
Here's some hastily fixed up code vomited to the page:
```js
const container = {
  _store: {},
  _miss(dep) {
    throw new Error(`Dependency not ready: ${dep}`)
  },
  get(dep) {
    return this._store[dep] || this._miss(dep)
  },
  provide(dep, as) {
    this._store[dep] = as
  }
}

function inject(...deps) {
  return fn => (...args) => fn(...args, ...deps.map(dep => container.get(dep)))
}

const registerDI = {
  onInit() {
    for (const property in this) {
      container.provide(property, this[property])
    }
  }
}

init({
  plugins: [
    registerDI
  ]
})
```

The Rematch plugins already mostly use a proxy in their `expose`. `dispatch`, `getState`, etc. work this way. I could add a getter that always returns callable functions. With, the annotation and the return type are reflective.

d3dc:
I mentioned this in #208, but 90% of the time, the lazy resolution of modules should work without a container.

I can often just import directly with the caveat that I can't call the imports outside of the model's members
```js
import { getState, select} from '@/store'
```

hardfist:
> Thanks @d3dc for all the work. I worry that this may add unnecessary complication to the API at this time. Will close as wontfix

without the support of `extraArguments` it can be hard to support universal rendering using rematch, which need inject http_client into effects to call service.

WebdevMerlion:
So. What about extending effects factory by `select`? Due to circular dependencies its not good to import store directly in model. Or should i use native redux selectors to properly work with state in effects? or `inject` as showed above?

PR #468, opened 2018-07-09
Update the npm install guide for @rematch/immer

2hu12:
The default version installed by `npm install @rematch/immer` was `0.1.0`, and it would get an error like [issues349](https://github.com/rematch/rematch/issues/349) with `rematch@next` installed.


ShMcK:
Thanks :)

PR #484, opened 2018-07-25
fix Model types: extend properly from ModelConfig

danielpa9708:


ShMcK:
Thanks daniel :) Will publish with the next release.

Issue #483, opened 2018-07-24
Cannot use selector plugin with "slice"


danawoodman:
If I have a model with the following selectors:

```js
  selectors: {
    double: () => s => s.a * 2,
  },
```

and I log out:

```js
import { select } from '@rematch/select'

console.log(select)
```

I see:

```js
{ foo: { double: [function] } }
```

As I would expect. If I instead try to use the `slice` as outlined in the docs like:

```js
  selectors: slice => ({
    double: () => slice(c => c * 2),
  }),
```

I see:

```js
{ foo: {} }
```

It appears passing in `selectors` as a function causes the selectors not to get created. 

Am I missing something or is this a bug?

Versions:

```json    "@rematch/core": "1.0.0-beta.5",
    "@rematch/immer": "1.0.2",
    "@rematch/select": "1.0.2",```

Thanks!

**Comment by d3dc** [comment]
Hey @danawoodman

Its awesome to see you're using  the new API, but it doesn't look like there is a published release of 2.0.0. 

Last I checked, it was running into typing issues with the namespace extension I did when it went to be packaged. I can definitely put some time into figuring it out.

**Comment by danawoodman** [comment]
Hmm, I read on issues #309 [here](https://github.com/rematch/rematch/issues/309#issuecomment-404062032) that these were published and this was how to use the API. I've also tried with `getSelect` but get the same results.

**Comment by danawoodman** [comment]
@d3dc I'm not sure what to do at this point. Should I be using a particular version? How can I get selectors working with 1.0.0-beta.5? Thanks

**Comment by d3dc** [comment]
This should be the docs for v1 of the plugin:
https://github.com/rematch/rematch/blob/af82445b22d50746d4e68dbdc1e68a81209d7de9/plugins/select/README.md

v1 of Rematch will hopefully ship with v2 of the plugin, so my link in the select plugin docs doesn't quite make sense.

**Comment by danawoodman** [comment]
I'm still confused... In those docs, it says to do:

```js
import { select } from '@rematch/select'
```

Wouldn't you want to do this instead:

```js
import { select } from './store'
```

Also, should I follow that guide if I plan on using v1 or is that going to change once v1 is out? Is anyone using v1 beta and select successfully together?

d3dc:
@danawoodman I'm putting in a bit of a stopgap PR that might let us publish the right version.

IMO, you shouldn't use the plugin before the breaking changes, just you can't currently install that version. I'm sorry this is all pretty confusing.

[...]

PR #488, opened 2018-07-26
Attempt to fix typings

d3dc:
Since it seems like I've created a snafu with #483, I massaged the typings for the two plugins - `select` and `typed-state` that extend `Rematch.ModelConfig`

[I tried to do it according to the plugin recipe in the docs](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-plugin-d-ts.html).

I also removed `subscriptions`


d3dc:
I'm able to get `npm run build` to run with an ambient `.d.ts` for `@rematch/select` when mapping `@rematch/core` to the local path.

The local path mapping is necessary because the published version of `core` has duplicate typings.

ShMcK:
@d3dc I've added you as a collaborator for rematch. I hope that makes your life easier :)

d3dc:
@ShMcK I'll see if it helps. I can fix the link in my docs now 😉.

The trick to everything was doing:
```
// a/plugins/select/tsconfig.json
 {
   "compilerOptions": {
+      "baseUrl": "./src",
       "outDir": "./lib",
       "sourceMap": true,
       "target": "es5",
       "lib": ["es2017"],
       "moduleResolution": "node",
+      "paths": {
+        "@rematch/core": ["../../../"]
+      },
   },
```

This fixes the `declare module '@rematch/core'` part and the import in the code.

I'm assuming this is because the published (`node_modules`) version does not have the typings. I guess I go ahead and try to publish?

d3dc:
Typed-state is still not building even when I replace the build config...

ShMcK:
Sorry, I guess we've both gone down this rabbit hole. 

@d3dc thanks for joining as a contributor. Feel free to manage and merge in any PRs related to the select plugin. I appreciate the help :)

ShMcK:
I'm still getting an error on build.

```
[!] (rpt2 plugin) Error: rematch/plugins/select/src/index.ts(95,8): semantic error TS2349 Cannot invoke an expression whose type lacks a call signature. Type 'never' has no compatible call signatures.
```

In reference to the following typings:

```
	export interface ModelConfig<S = any, SS = S> {
		selectors?: ModelSelectorsConfig<S>,
	}
```

How did you manage to run it with just a warning?

ShMcK:
@d3dc - would you like to be added as an npm maintainer for `@rematch/select`? If so, just send me an email with your npm username. My email should be publicly available on my Github.

d3dc:
I think I might be wrong about why my changes get `typed-state` to compile. I Removed the `ModelConfig` interface as a sanity check, and it still built.

ShMcK:
I finally managed to get it to build by changing the reference in the select plugin to the local build.

It should be published now as 2.0.0. Thanks again for all of your effort, this was not easy.

And again, let me know if you'd like more shared control over publishing the select plugin.

Issue #489, opened 2018-07-28
immer plugin should have 'blacklist' config to support cooperation with loading plugin


linonetwo:
There is  a 'blacklist' config field in loading plugin https://github.com/rematch/rematch/blob/master/plugins/loading/src/index.ts#L5

But immer plugin don't have one https://github.com/rematch/rematch/blob/master/plugins/immer/src/index.ts#L25

To support https://github.com/rematch/rematch/issues/345, there should be a 'blacklist' config field in immer plugin.

linonetwo:
How about wrap https://github.com/rematch/rematch/blob/master/plugins/immer/src/index.ts#L10 with
```js
// ignore items not in whitelist
if (!config.whitelist || !config.whitelist.includes(key)) {
  ...
}
```
?

Issue #345, opened 2018-04-20
@rematch/loading have trouble with @rematch/immer


pciapcib:
Hello, thanks for amazing rematch and @ShMcK .

I have trouble in using rematch plugins. I found that @rematch/loading and @rematch/immer have conflict when I use both in a project. The strange behavior is that @rematch/loading is always false(0.3.6) or 1(>= 0.4.0) and @rematch/immer is showing empty action in redux devtools.

I am using parcel-bundle for my tiny project, I hope it's not the reason of this issue...

``` js
// store.js
import { init } from '@rematch/core'
import immerPlugin from '@rematch/immer'
import loadingPlugin from '@rematch/loading'
import { routerMiddleware, routerReducer as routing } from 'react-router-redux'

// via babel-plugin-glob-import
import * as models from './models/*.js'

import { history } from './router'

const immer = immerPlugin()
const loading = loadingPlugin()

const store = init({
  plugins: [immer, loading],
  models,
  redux: {
    middlewares: [routerMiddleware(history)],
    reducers: {
      routing
    }
  }
})

export default store
```

ShMcK:
Perhaps an immer `whitelist` or `blacklist` options may help. 

```js
immerPlugin({
  blacklist: ['loading']
})
```

Any better ideas?

linonetwo:
I think this solution is enough, just one line of config. But it needs to be mentioned in loading plugin's documentation.

linonetwo:
Even use [immutable-js-example](https://github.com/rematch/rematch/blob/master/plugins/loading/README.md#immutable-js-example) is impossible:

https://stackblitz.com/edit/react-ussyfk?file=store.js

linonetwo:
@jackwilsdon  Yeah, loading plugin is using copyReducer:
https://github.com/rematch/rematch/blob/faf9db021b843866620da7a8dff4028124f03e68/plugins/loading/src/index.ts#L24-L38
And immer plugin produce it:
https://github.com/rematch/rematch/blob/faf9db021b843866620da7a8dff4028124f03e68/plugins/immer/src/index.ts#L10-L18
The only difference is `cntState`, which is outside of `produce`, but it don't seem to be the problem.

linonetwo:
![wx20180729-164336](https://user-images.githubusercontent.com/3746270/43364539-9727cd2a-934e-11e8-875b-2d80fdcde103.png)

Result returned from loading plugin is correct. It must be immer plugin's fault.

[...]

PR #490, opened 2018-07-28
add whitelist and blacklist to immer plugin

linonetwo:
Try to fix https://github.com/rematch/rematch/issues/489 and https://github.com/rematch/rematch/issues/345 .

### DIFF
--- a/plugins/immer/src/index.ts
+++ b/plugins/immer/src/index.ts
@@ -2,32 +2,62 @@ import { Models, Plugin } from '@rematch/core'
 import produce from 'immer'
 import { combineReducers, ReducersMapObject } from 'redux'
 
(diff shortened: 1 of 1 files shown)


jackwilsdon:
:-1: on this, it shouldn't be required as immer supports both copy reducers (i.e. traditional style) and mutating reducers. We should aim to fix the issue with `@rematch/immer` instead.

linonetwo:
@jackwilsdon  Agree with you, I'll try to found out why.

jackwilsdon:
Glad to hear it! Thanks for looking into it some more. It is a pain debugging the compiled code, I found it to be one of the bigger issues when I was doing some work on the library.

Cheers! :tada:

PR #491, opened 2018-07-29
Fix immer plugin's cooperation with loading plugin

linonetwo:
Fixs https://github.com/rematch/rematch/issues/345

Much simplier than https://github.com/rematch/rematch/pull/490

### DIFF
--- a/plugins/immer/src/index.ts
+++ b/plugins/immer/src/index.ts
@@ -10,7 +10,7 @@ function combineReducersWithImmer(reducers: ReducersMapObject) {
 		reducersWithImmer[key] = (state, payload) => {
 			if (typeof state === 'object') {
 				return produce(state, (draft: Models) => {
-					reducerFn(draft, payload)
+					return reducerFn(draft, payload)
 				})
 			} else {
 				return reducerFn(state, payload)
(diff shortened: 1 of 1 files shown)


jackwilsdon:
The only issue I can see with this is that it will break code like this (which is wrong anyway I guess);

```JavaScript
init({
  plugins: [immer],
  models: {
    count: {
      state: {
        value: 0,
      },
      reducers: {
        increment(state, payload) {
          state.value += 1;
          return state;
        },
        decrement(state, payload) {
          state.value -= 1;
          return state;
        },
      },
    },
  },
});
```

This will now break because immer throws an error if a reducer both modifies the state and returns a value;

https://github.com/mweststrate/immer/blob/979d7a34dd0331b5ba44983bf55cc545ad9ea2f4/src/proxy.js#L155-L157

Although I'm not sure how big of an issue this is.

jackwilsdon:
Scratch that comment I just left (and hastily deleted), it turns out if you return the proxy from `immer` then everything is fine anyway. I guess it would still break in cases like this?

```JavaScript
const model = {
  state: {
    value: 0,
  },
  reducers: {
    increment: state => state.value += 1,
  },
}
```

linonetwo:
@jackwilsdon Seems immer's developers have handled it properly, 👍 for them. Try https://stackblitz.com/edit/react-ksr6xw?file=index.js

linonetwo:
OK, `Create the next immutable state by mutating the current one` if the next state has the same type as the previous state.
```js
if (typeof state === 'object') {
	return produce(state, (draft) => {
          const nextState = reducerFn(draft, payload)
          if (typeof nextState === 'object') {
            return nextState;
          }
        })
}
```

ShMcK:
Looks good, will publish later today. :)

Issue #390, opened 2018-05-28
Complex examples


metakermit:
I like the framework concept. Are there perhaps any examples out there of actions / effects spread out across multiple files? It would be nice if the framework supported some way to *register* actions, effects or even parts of the model in different parts of the codebase. For example:

```js
// myReducer.js
import {reducer} from '@rematch/core'
function myReducer(state, payload) { ... }
// register this reducer and somehow bind it to the state
reducer(myReducer)
```

This way it wouldn't be necessary to hand-collect everything manually into a giant object. Of course, it would be possible to create something like this manually in every project, but if the goal is to reduce boilerplate, I think it would be nice to offer some "opinionated" way to easily manage larger projects.

BTW, I can't get to the examples [in the docs](https://rematch.gitbooks.io/rematch/#examples). The url always takes me to the table of contents.

williamconnolly:
Yeah, I'm absolutely loving how this made Redux easy to integrate. However, it took probably just about as much time to piece together the differences between typescript examples and the react example (that still imports "rematch-x", I guess an outdated name. 

Keep doing what y'all are doing, you're making a kick ass module and I hope I can contribute some day as I get more versed in frontend.

ShMcK:
Thanks @metakermit - I've put a fix in for the examples link.

And @williamconnolly - I've also updated the count README, which was outdated.

ShMcK:
In reference to importing reducers, it might to helpful to note:

- You can trigger reducers/effects in other models
- To reuse reducers or effects you can spread them. They are really just objects

```js
const commonReducers = {
  set(state, payload) {
    return payload
  },
}

const otherModel = {
  reducers: {
    ...commonReducers,
  }
}
```

williamconnolly:
@ShMcK in your updated example, you follow a similar pattern as elsewhere in the docs, with mapDispatchToProps. Is this strictly necessary? 

Since dispatch.modelName.reducerName(...) can be called from anywhere in the code, what's the purpose of providing those methods as props to the main Component?

I've been following the pattern, but I have a subcomponent of my main one, and have been passing IT props with those dispatch methods. Why not simplify it and have every dispatch-interaction be a global dispatch.model.reducer(...) call?

ShMcK:
@williamconnolly I think there's danger in globals, especially when you try to scale your project. For example, you may want to use multiple stores for both organization & performance reasons.

I would consider the global dispatch to be a last resort / hack to get things done.

williamconnolly:
[...]

PR #500, opened 2018-08-10
Select examples

### LINKED ISSUE #298: Cross-model actions/reducers?
I like what this library is trying to achieve. There is one thing that I couldn't grasp looking at the documentation / examples and code...

Say you have 2 models: A and B. You have some action defined in model A and you want a reducer in model B to listen to that action to mutate model's B state. How do you accomplish this?

This is useful for badges on the taskbar, handling auth actions (login / logout) where you want to clear and mutate state of non-auth related models, etc.

d3dc:
This updates the examples to use `@rematch/select` 2.0.0.

closes #298

I added a selector to the react/count example since it shows `connect`. I can take that commit out if it should remain simple. (or make a new example)

### DIFF
--- a/examples/ts/count/package.json
+++ b/examples/ts/count/package.json
@@ -1,6 +1,6 @@
 {
   "name": "rematch-typescript-example",
-  "version": "1.0.0-beta.3",
+  "version": "1.0.0-beta.5",
   "private": true,
   "dependencies": {
     "@rematch/core": "../../..",
--- a/examples/ts/count/src/models/sharks.ts
+++ b/examples/ts/count/src/models/sharks.ts
@@ -17,9 +17,9 @@ export const sharks = createModel({
 			dispatch.sharks.increment(payload || 1)
 		},
 	}),
-	selectors: {
-		total(state: SharksState) {
-			return state
+	selectors: slice => ({
+		total() {
+			return slice
 		},
-	},
+	}),
 })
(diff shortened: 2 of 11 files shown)


ShMcK:
I think it may be easier for now to keep the count example as simple as possible. The idea of selectors can even be a bit of learning curve for those new to functional programming.

What do you think about moving a count-select example under `plugins/select/examples`?

d3dc:
I went ahead and made a new example under select.
Should the ts example still have a selector then?

ShMcK:
That's a good question. 

I might consider that the basic JS & TS examples to serve different purposes.

JS: shows how to make a basic app.
TS: shows an example of all kinds of typings. 

What are your thoughts?

d3dc:
I don't think I'm the right crowd to have an opinion. I think I agree with you that TS is more about showing how types are dealt with. An example of a working TS project is one where it compiles with hints.

I have been thinking about the todo example I made in a codesandbox, and wondering if it might adapt well for #390; but thats not here nor there for this chain of issues.

ShMcK:
Your todos complex example may be better linked to in a codesandbox on the README under the examples section. 

Let me know when you think this PR is ready :)

d3dc:
I just tested running it. I guess I thought it wasn't obvious that you could remove things. If the example is sound then, it should be good.

PR #501, opened 2018-08-13
fix doc

onlyling:
store.model

store.getState()
{ count: 0, countB: state: 99 }  -> { count: 0, countB: 99 }


ShMcK:
Much thanks :)

PR #506, opened 2018-08-21
Add another, fuller example of combining selectors with raw state

danawoodman:


plugins/select/README.md
```diff
@@ -200,6 +198,24 @@ connect(select(models => {
 }))(...)
 ```
 
+Here is a full example combining raw, uncomputed state with selected state:
+
+```js
+const mapStateToProps = state => {
+  const selection = select(models => ({
```
d3dc: You should create the selector outside of mapStateToProps. 

You'll get a lot more performance if your selector is static and the cache is re-used.

plugins/select/README.md
```diff
@@ -200,6 +198,24 @@ connect(select(models => {
 }))(...)
 ```
 
+Here is a full example combining raw, uncomputed state with selected state:
```
d3dc: Maybe lead into this with motivation:

"You might want to combine raw, uncomputed state with your selectors. In this case, the combined selector from select can be called directly"

I'll have to fix some of the language here around select itself later, so maybe just leave it for that.

Issue #515, opened 2018-09-08
Effect function with multiple arguments and state


kkshyu:
Related to this issue: #267 
i need the state and the multiple arguments in effects, and i followed the example from [here](https://github.com/rematch/rematch/blob/master/examples/ts/count/src/App.tsx).

the type cannot match in the multiple arguments, but it works in the single argument.
```
[ts]
Argument of type '{ lectureId: string; chapters: IChapter[]; }' is not assignable to parameter of type 'Action<{ lectureId: string; chapters: IChapter[]; }, void>'.
  Object literal may only specify known properties, and 'lectureId' does not exist in type 'Action<{ lectureId: string; chapters: IChapter[]; }, void>'.
```

here are the codes:
// models.ts
![image](https://user-images.githubusercontent.com/8722165/45252932-bd5bae00-b35e-11e8-86c4-97820104746f.png)
// component.ts
![image](https://user-images.githubusercontent.com/8722165/45252961-0e6ba200-b35f-11e8-9d6e-751d70fe13b0.png)
![image](https://user-images.githubusercontent.com/8722165/45252962-10356580-b35f-11e8-811e-215dc1b12848.png)

In `component.ts:36`, the multiple arguments is not working, but in `components.ts:43` the single argument is working. Any wrong configure?

kkshyu:
@ShMcK i just solved this and sent the PR, plz review it can give some feedbacks. thx!

ShMcK:
Issue should be resolved with @rematch/core@1.0.2. Thanks @kkshyu

FirsovRG:
@Krasnov8953, I have the same problem too

sup4x:
Guys, I have the same problem as guys have @Krasnov8953 , @FirsovRG

ShMcK:
Happy to review any PRs that resolve the issue

PR #521, opened 2018-09-22
Fix typings issue for effects that use rootState

koss-lebedev:
Hi guys,

There is an issue with type inference when effects access `rootState`. This is the same issue that's described here: https://github.com/rematch/rematch/issues/515

So type inferrence for this snippet will work:

```ts
const model = createModel({
  /* omitted */
  effects: dispatch => ({
    async loadAsync(userId: string) {
      /* omitted */
    },
  }),
})
```

But it will break for this one since we use `rootState`

```ts
const model = createModel({
  /* omitted */
  effects: dispatch => ({
    async loadAsync(userId: string, rootState) {
      /* omitted */
    },
  }),
})
```

The problem is in this declaration on the second to last line:

```ts
export type ExtractRematchDispatcherAsyncFromEffect<E> =
  E extends () => Promise<void> ? RematchDispatcherAsync<void, void> :
  E extends (payload: infer P) => Promise<void> ? RematchDispatcherAsync<P, void> :
  E extends (payload: infer P, meta: infer M) => Promise<void> ? RematchDispatcherAsync<P, M> :
  RematchDispatcherAsync<any, any>
```

When we define effect function in the model, second argument that is called `meta` here will always be the `rootState`. However, the effect function that we want to call from inside our components must accept only one argument, which is a `payload`. The problem is that we pass inferred `M` type to `RematchDispatcherAsync`, so the effect function that we want to call from inside our components will expect the second argument of type `M` to be passed, which is incorrect.

This PR resolves this issue.

### DIFF
--- a/src/typings.d.ts
+++ b/src/typings.d.ts
@@ -17,7 +17,7 @@ export type RematchRootState<M extends Models> = ExtractRematchStateFromModels<M
 export type ExtractRematchDispatcherAsyncFromEffect<E> =
   E extends () => Promise<void> ? RematchDispatcherAsync<void, void> :
   E extends (payload: infer P) => Promise<void> ? RematchDispatcherAsync<P, void> :
-  E extends (payload: infer P, meta: infer M) => Promise<void> ? RematchDispatcherAsync<P, M> :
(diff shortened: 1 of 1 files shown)


ShMcK:
Thanks @koss-lebedev. Will merge and release soon.

Issue #526, opened 2018-09-29
typescript: how to get `dispatch.model.effect` to show correct type signatures


derrickbeining:
I've been struggling for a while trying to figure out how to set up `rematch` for my Typescript/React project. I followed the [TypeScript recipe in the docs](https://rematch.gitbooks.io/rematch/docs/recipes/typescript.html) and looked through [examples/ts/count](https://github.com/rematch/rematch/tree/master/examples/ts/count) (which isn't compiling because of `selectors` on the `shark` model).  

In particular, I can't figure out how to get `dispatch.<model>.<effect>` to give me the correct function signature for my effects. For instance:

Here's where I defined `dispatch.api.getAllPosts`:

![screen shot 2018-09-29 at 3 06 19 pm](https://user-images.githubusercontent.com/25019711/46249890-c50eef80-c3fe-11e8-9e93-627ad585218b.png)

But when I try to use it, I get this error:

![screen shot 2018-09-29 at 3 23 41 pm](https://user-images.githubusercontent.com/25019711/46249907-29ca4a00-c3ff-11e8-82ff-d8fb3625d52f.png)

Anyone know what I'm doing wrong?

derrickbeining:
looks like this is the same as #478 and #520

derrickbeining:
closing since duplicate

derrickbeining:
Reopened because #478 is closed.

derrickbeining:
See my comments [here](https://github.com/rematch/rematch/issues/478#issuecomment-425691702) and my possible solution [here](https://github.com/rematch/rematch/issues/478#issuecomment-425692278)

ShMcK:
I believe the issue should be resolved in @rematch/core@1.0.2. Thanks @kkshyu

Issue #520, opened 2018-09-21
TypeScript: Synchronous effects' parameters requires Action<any, void>


zsxsoft:
Same issue with: https://github.com/rematch/rematch/issues/478

Reproduce: 

1. Add a synchronous effect in https://github.com/rematch/rematch/blob/master/examples/ts/count/src/models/dolphins.ts.

```typescript
	effects: dispatch => ({
		aaa (s: string) {
			// do nothing
		},
		async incrementAsync(a: number) {
			await delay(500)
			dispatch.dolphins.increment()
		},
	}),
```

2. Call it:
![image](https://user-images.githubusercontent.com/2398785/45866781-6e793400-bdb4-11e8-9724-02d0ed4e15be.png)

TypeScript returns:
```
[ts] Argument of type '111' is not assignable to parameter of type 'Action<any, void>'.

ShMcK:
Issue should be resolved with @rematch/core@1.0.2. Thanks @kkshyu

raitucarp:
this issue still exist on @rematch/core@1.0.6

Issue #478, opened 2018-07-18
typescript effects need a action as a parameter


cnyballk:
Hi. I need some help. I introduced typescript in react and rematch, but I ran into some problems that looked like this.
![image](https://user-images.githubusercontent.com/25214319/42864508-4c6e445c-8a99-11e8-8aa7-25ad4221e524.png)
In the definition,
![image](https://user-images.githubusercontent.com/25214319/42864532-63a18e4a-8a99-11e8-8453-0f5f3471055e.png)
It reminds me that I need a action as a parameter, but I'm pretty clear I don't need one
![image](https://user-images.githubusercontent.com/25214319/42864559-7425c6be-8a99-11e8-9e89-35478de41502.png)
Now I don't know what I'm gonna do. I don't want to use as any. Can you help me?

cnyballk:
This is my environment.
 "@rematch/core": "^1.0.0-beta.4",

ShMcK:
In the case of `models` are you handling them in a similar way to this link?

https://github.com/rematch/rematch/blob/master/examples/ts/count/src/store.ts#L7

cnyballk:
Yes, I did.

Izhaki:
Running into the same issue with:

```json
  "dependencies": {
    "@rematch/core": "^1.0.0-beta.4"
  },
  "devDependencies": {
    "typescript": "3.0.3"
  }
```

models.js:

```typescript
import { createModel } from '@rematch/core';

export const auth = createModel({
  state: 0, // initial state
  effects: dispatch => ({
    userAlreadyLoggedIn() {
      console.log('SIM: userAlreadyLoggedIn');
    },
  }),
});
```

index.js:

```typescript
import { init, RematchDispatch } from '@rematch/core';
import * as models from './models';

type IModels = typeof models;

const store = init({
  models,
});

const dispatch: RematchDispatch<IModels> = store.dispatch;

dispatch.auth.userAlreadyLoggedIn();
// [ts] Expected 1 arguments, but got 0.
// (method) userAlreadyLoggedIn(action: Action<any, any>): Promise<Dispatch<Action<any, any>>> (+1 overload)

```

derrickbeining:
I'm having the same problem.

```
  "dependencies": {
    "@rematch/core": "^1.0.0-beta.5"
  },
  "devDependencies": {
    "typescript": "3.1.1"
  }
```

[...]

Issue #398, opened 2018-06-05
[TypeScript] error when dispatch action using `dispatch.<model>.<effect/reducerName>`


kebot:
![image](https://user-images.githubusercontent.com/289392/41005663-bbf229de-691e-11e8-8e3c-1511f97f79b7.png)

![image](https://user-images.githubusercontent.com/289392/41005864-81034b72-691f-11e8-85b4-10a126ed462b.png)

I have define types to the `effects` function

![image](https://user-images.githubusercontent.com/289392/41005910-b2105584-691f-11e8-83b5-368f70fcd413.png)

A workaround would be always use `dispatch({type: 'model/dispatch'})`, but will lost all the typescript benefit.

ShMcK:
Two things may help:

1. Use alpha 7

2. Checkout the [TS example](https://github.com/rematch/rematch/tree/master/examples/ts/count), in particular notes these [lines](https://github.com/rematch/rematch/blob/master/examples/ts/count/src/App.tsx#L12-L18). Also note how `models` are given a type within "store.js". Also note how the `createModel` is used, and that the `payload` is typed within models/sharks.ts.

If you still have a problem, let us know :)

kebot:
Just play the latest ALPHA this weekend.

Another thing I just found, in order to get type suggesting in Model's reducers, you need assign it to a Model type variable.

```typescript
import { createModel, Model } from '@rematch/core'

interface State = {
  isFetching: bool,
  todos: Array<string>
}

const todosModel : Model<State>= {
  state: {isFetching: false, todos: []},
  reducers: {
    fetchSuccess(state, todos) {
      >>> you get right type suggesting of `state` here
    }
  }
}

export default createModel<State>(todosModel)
```

ShMcK:
Definitely we could use some improved TS documentation. It's not obvious.

brunolemos:
@kebot you will lose autocomplete for the `todosModel` if you use `: Model`. You can do `fetchSuccess(state: State` instead.

PR #529, opened 2018-09-30
fix typings in effects

kkshyu:
return type in `effects` can be not only `void`

sometimes we need the return value from `effects`, so we will code like this:
```
// models/sample.ts
{
    effects: {
        async noPayloadWithReturnAsync() {
          return {
            res: 'noPayloadWithReturn'
          }
       },
   }
} 
```
```
// somewhere.ts
dispatch.sample.noPayloadWithReturnAsync()
.then(ret => {
    // do something with `ret.res`
})
```

in this PR, i allow the return type in `effects` not only `void`. 
The original will throw type error if you have return types except for `void`

there are bunch of issues about this: #398 #478 #520 #526 #515 
and i have tested all the cases on them

image

image


ShMcK:
Thanks @kkshyu! We can always use more help with typings

Issue #522, opened 2018-09-26
TypeScript: boolean parameters requires Action<boolean, void>


bagr001:
Hi, I found an inssue with typescript and boolean action parameters. It occurs both in reducers and effects. Interesting is that the issue is only with boolean types because with other types (such as strings, etc...) it works nice. I went thru the rematch typings but I have no clue what the problem might be :-/ Any idea? Thanks.

![image](https://user-images.githubusercontent.com/5631971/46080505-65180f00-c19a-11e8-9b11-b9c1b6bfc49b.png)

bagr001:
@ShMcK Maybe it is a TS bug, but I probably found fix for this. Just add exact typing for this case and it seem to work.

```ts
P extends boolean ? ((payload: boolean) => Action<boolean, any>) :
```
```ts
P extends boolean ? ((payload: boolean) => Promise<Action<boolean, any>>) :
```

```ts
export type RematchDispatcher<P = void, M = void> =
  ((action: Action<P, M>) => Redux.Dispatch<Action<P, M>>) &
  ((action: Action<P, void>) => Redux.Dispatch<Action<P, void>>) &
  (
    P extends boolean ? ((payload: boolean) => Action<boolean, any>) :
    P extends void ? ((...args: any[]) => Action<any, any>) :
    M extends void ? ((payload: P) => Action<P, void>) :
    (payload: P, meta: M) => Action<P, M>)

export type RematchDispatcherAsync<P = void, M = void> =
  ((action: Action<P, M>) => Promise<Redux.Dispatch<Action<P, M>>>) &
  ((action: Action<P, void>) => Promise<Redux.Dispatch<Action<P, void>>>) &
  (
    P extends boolean ? ((payload: boolean) => Promise<Action<boolean, any>>) :
    P extends void ? ((...args: any[]) => Promise<Action<any, any>>) :
    M extends void ? ((payload: P) => Promise<Action<P, void>>) :
    (payload: P, meta: M) => Promise<Action<P, M>>)
```

ShMcK:
Does this typings issue persist with @rematch/core v1.0.2 ?
If so, happy to merge your fix in as a PR.

ShMcK:
Will be resolved in @rematch/core @1.0.5

PR #536, opened 2018-10-03
fix typings - boolean Action parameters requires Action<boolean, void>

bagr001:
Opening PR for #522


ShMcK:
Would changing the types to `any` also work? It may cover more use cases.

bagr001:
@ShMcK It would also work, but type-hint will always show 
`(property) setFoo: (...args: any[]) => Action<any, any> (+2 overloads)`
instead of proper type
`(property) setProjectExists: (payload: boolean) => Action<boolean, any> (+2 overloads)`

Javi:
I think this is happening again on Rematch Core v1.2.0 when calling a reducer with a boolean parameter.

Issue #524, opened 2018-09-28
TypeScript: Select plugin issue


nWidart:
Hello,

In a typescript project, I'm trying to use the select plugin, with the re-select api.

In the Typescript example, there's this: https://github.com/rematch/rematch/blob/master/examples/ts/count/src/models/sharks.ts#L20
```js
	selectors: slice => ({
		total() {
			return slice
		},
	}),
```

However, if I use the same structure I'm getting the following error:

```
error TS2717: Subsequent property declarations must have the same type.  Property 'selectors' must be of type '{ [key: string]: (state: SS, ...args: any[]) => any; }', but here has type 'ModelSelectorsConfig<S>'.
```

It looks like the module overwrite, https://github.com/rematch/rematch/blob/master/plugins/select/src/typings.d.ts#L68, wasn't taken into account?

**This happens when requiring the rematch/select plugin, even without any personal code.**

***

Also, when trying to use my selector, I'm getting this:

Selector (not using re-select):

```js
  selectors: {
    allEnabledEngagementPartners() {
      return (rootState) => rootState.data
    },
  },
```

```js
TS2349: Cannot invoke an expression whose type lacks a call signature. Type 'Selector<any, any, any>' has no compatible call signatures.
```

***

Did I miss something?

Thanks!

d3dc:
This is just `core` typings taking precedence. My attempt to fix it is in master. I'm pretty sure we're just waiting on the next release ( fixing all the other typescript :) )

ShMcK:
Typings seem to work in the current release, with the exception of select plugin typings. 

Inside of the select plugin, it attempts to modify the core module typings by extending the Model with a `selectors` key.

```ts
declare module '@rematch/core' {
	export interface ModelConfig<S = any, SS = S> {
		selectors?: ModelSelectorsConfig<S>,
	}
}
```

TypeScript does not seem to register type extensions if they are located in a different module.

Any suggestions on how we might proceed?

d3dc:
@ShMcK 

> error TS2717: Subsequent property declarations ...

Doesn't this imply the compiler found the extension?

ShMcK:
True, the compiler seems to find the type extension. but still creates conflicting typings when using `selectors`.

Unfortunately, I'm not sure what causes the problem, even after reading the TS docs under [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html).

Any ideas?

d3dc:
[...]

PR #539, opened 2018-10-08
Fix `Slicer` type and export select plugin types

19majkel94:
This PR fix the `Slicer` type signature (from Union `|` to Intersection `&`) and the type of `RematchStore` select property (missing `RootState` type in generic).

It also exports all the types as `@rematch/core` for further integration and own types extensions by consumers.

### DIFF
--- a/plugins/select/src/typings.d.ts
+++ b/plugins/select/src/typings.d.ts
@@ -13,57 +13,54 @@ export interface SelectConfig {
 declare const createSelectPlugin: (config?: SelectConfig) => R.Plugin<R.Models, R.Action<any, any>>
 export default createSelectPlugin
 
-type Selector<S, P = any, R = any> =
+export type Selector<S, P = any, R = any> =
 	Reselect.Selector<S, R>
 	& Reselect.ParametricSelector<S, P, R>
 
-interface ModelSelectors<S> {
+export interface ModelSelectors<S> {
 	[key: string]: Selector<S>
 }
 
-interface StoreSelectors<S> {
+export interface StoreSelectors<S> {
 	[key: string]: ModelSelectors<S>
 }
 
-type SelectorFactory<S, P = any, R = any> =
+export type SelectorFactory<S, P = any, R = any> =
 	(this: ModelSelectors<S>, models: StoreSelectors<S>) => Selector<S, P, R>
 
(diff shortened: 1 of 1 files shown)


19majkel94:
I would also like to propose another change - improving model typings. The `createModel` signature is not checking the model config param itself, it has to be done in two steps:
```ts
export function createModel<TState>() {
  return <T extends ModelConfig<TState>>(model: T): ModelWithState<TState, T> =>
    model;
}
// then
const cart = createModel<CartState>()({
  // ...
});
```

The `ModelWithState` type will also improve usage of `RematchRootState` which reads the `state` property of model, so when someone provides `[]` it's inferred to be `never[]`, or when state type is  `T | null` and the `state` value is `null`:
```ts
export type ModelWithState<TState, T extends ModelConfig<TState>> = Pick<
  T,
  Exclude<keyof T, "state">
> & { state: TState };
```

But this might be a breaking change and I can use it in my project without any conflicts.

ShMcK:
Thanks @19majkel94! I'm all for simplifying the typings.

If it's a breaking change, we'll have to carefully update all the typings in a single PR. Feel free to open an issue, we can see if the topic gets more support.

d3dc:
@19majkel94 This looks great! Being not great with typescript, I had a few questions. 

Is the idea to just export every type? Things like `Selector` just remove type information from `Reselect.Selector` (`createSelector` returns a `Reselect.Selector`, no?), while things like `ModelSelectors` are more convenience. I guess no harm if they are available to extend? How does intellisense interact with the `Selector` type?

19majkel94:
> Is the idea to just export every type?

All types are exported in rematch core. It allows to write custom types when the built one fails (like now https://github.com/rematch/rematch/issues/524#issuecomment-427843877) as copy-pasting is not always a good solution.

> I guess no harm if they are available to extend? How does intellisense interact with the Selector type?

It will just show you two types proposition with info about the package from which it will import it. Thanks to types import there's no name collision.

d3dc:
@19majkel94 I guess I'm wondering if we should be able to infer `Reselect.ParametricSelector` vs. `Reselect.Selector`. I'm giving this a go in a codesandbox right now, so I might've missed it.

PR #541, opened 2018-10-09
fix persist plugin declaration file path

### LINKED ISSUE #518: persist plugin typescript path
index.d.ts should be
```ts
import { Plugin } from '@rematch/core';
export declare const getPersistor: () => any;
declare const reactNavigationPlugin: (
  config?: {}
) => Plugin<
  import('../../core/src/typings').Models,
  import('../../core/src/typings').Action<any, any>
>;
export default reactNavigationPlugin;
```
but is this:
```ts
import { Plugin } from '@rematch/core';
export declare const getPersistor: () => any;
declare const reactNavigationPlugin: (
  config?: {}
) => Plugin<
  import('../node_modules/@rematch/core/src/typings').Models,
  import('../node_modules/@rematch/core/src/typings').Action<any, any>
>;
export default reactNavigationPlugin;
```

RunningCoderLee:
fix #518

### DIFF
--- a/plugins/persist/package-lock.json
+++ b/plugins/persist/package-lock.json
@@ -94,6 +94,12 @@
         "repeat-element": "^1.1.2"
       }
     },
+    "builtin-modules": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/builtin-modules/-/builtin-modules-2.0.0.tgz",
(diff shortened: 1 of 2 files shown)


ShMcK:
Sorry, I hadn't checked this PR earlier. It definitely would have resolved the issue faster. Really appreciate your quick fix.

Without seeing this, earlier this morning, I pushed and published a resolution. I'll try to be more on top of notifications in the future.

Issue #544, opened 2018-10-09
Reselect v4


19majkel94:
A few days ago, Reselect 4.00 had came to live:
https://github.com/reduxjs/reselect/blob/master/CHANGELOG.md#v400---20180930

The changes are rather small but it would be better if one of the plugin author could take a look at them and update this dep to the newest version. I've tried to run the tests locally but there's no test script in package.json nor `npx jest` doesn't work.

d3dc:
I never thought about it, but this actually strange: you can run all the plugin tests from the repo root. 

`npm test select` should run.

19majkel94:
@d3dc 😞 
```
PS C:\#Projekty\rematch> npm test select

> @rematch/core@1.0.6 test C:\#Projekty\rematch
> jest "select"

No tests found
In C:\#Projekty\rematch
  184 files checked.
  testMatch: **/__tests__/**/*.js?(x),**/?(*.)+(spec|test).js?(x) - 26 matches
  testPathIgnorePatterns: \\_book\\,\\lib\\,\\plugins,\\plugins\\react-navigation\\examples,\\plugins\\subscriptions,\\plugins\\loading-immutable,\\examples\\ts - 49 matches
Pattern: select - 0 matches
npm ERR! Test failed.  See above for more details.
PS C:\#Projekty\rematch>
```

d3dc:
Indeed. Testing plugins was broken by #528

19majkel94:
What about this? Is it worth to update?

PR #550, opened 2018-10-12
Add jest and config to select plugin

d3dc:
This just adds the ability to do `npm test` in `plugin/select`, since we can't re #544 - v1.0 of core removed running plugin tests.

I copy and pasted the same coverage configuration to the plugin config, but didn't do anything with it. Is there anything to do to include plugin coverage in Travis?

Is this the intended way to do plugin tests since the change?


ShMcK:
I suppose even if we return plugin coverage to core test coverage, it would still be wise to allow local tests. Will merge.

PR #554, opened 2018-10-22
Fix typings for select plugin

19majkel94:
I've rewritten the types from scratch, dividing into two sections - first is checking our config object and providing intelisense/type inference, the second one is for extracting proper shape of `select` from models.

It should cover all the use cases that I've found in docs and tests but I don't know how to create proper tests for type failing, like DefinitelyTyped has a dtslint suite. Maybe we can together figure out something for this 😉

### DIFF
--- a/plugins/select/src/typings.d.ts
+++ b/plugins/select/src/typings.d.ts
@@ -17,6 +17,10 @@ declare const createSelectPlugin: (
 ) => Rematch.Plugin<Rematch.Models, Rematch.Action<any, any>>;
 export default createSelectPlugin;
 
+export type Selector<TState, TReturns, TProps = void> = TProps extends void
+  ? (state: TState) => TReturns
+  : (state: TState, props: TProps) => TReturns;
+
 /****************************************
  *  types for selectors config in model *
  ****************************************/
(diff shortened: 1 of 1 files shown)


19majkel94:
Ping @d3dc as you wanted to see the fixed typings 😉

d3dc:
There's a lot here :)

d3dc:
So, what's holding me up right now is how coupled this is to Reselect. I'm still looking into the specific of the prop typings and maybe there is a reason you went this route.

My thinking though is: Do we need to define `Selector` at all?

If you use `createSelector` it has a proper return type. If you use a customized `createdSelector` hopefully it has a proper return type. If you return a vanilla function, it should still work.

This PR seems to omit the vanilla case and may potentially not interact with a custom configuration well.

19majkel94:
> So, what's holding me up right now is how coupled this is to Reselect. I'm still looking into the specific of the prop typings and maybe there is a reason you went this route.

It's just for convenience - I will replace it, leaving only OutputParametricSelector as in original typings.

d3dc:
I'm pretty sure jest can do some basic type checking from the existing tests - they just need to be written in typescript. I'm playing around with some typescript testing this weekend and will hopefully have some hint as to whether to roll this in.

Thanks for the prompt change!

ShMcK:
Any progress on this PR?

ShMcK:
Closing due to a lack of development. Feel free to re-open it if it needs revisiting.

19majkel94:
I'm using this typings in my apps and they works great. We only need someone who know the [`dtslint`](https://github.com/Microsoft/dtslint) and can create a tests for the API typings.

PR #563, opened 2018-10-31
semver: support redux patch versions

dcousens:
With `redux@4.0.1` out,  this package holds it back due to no `^`.


ShMcK:
Thanks @dcousens, hopefully we can trust redux to stick to semver :)

Issue #509, opened 2018-08-29
[QUESTION] How to use redux-undo with rematch?


Squeegy:
I'm having a doozy of a time trying to get rematch to play nice with [redux-undo](https://github.com/omnidan/redux-undo)

It seems to want you do `undoable(reducer)`, but it expects the reducer to be in redux standard format:

```js
(state, action) => newState
````

Where rematch uses this:

```
{
  [actionType]: (state, actionPayload) => newState
}
```

Is it possible to make this work? Is there a rematch specific undo package people are using?

**Comment by Squeegy** [comment]
That was a good hint!  So I think I'm on the right track with something like this?

```js
import { init } from '@rematch/core'
import { combineReducers } from 'redux'
import undoable from 'redux-undo'
import models from './models'

export const store: Store = init({
  models,
  redux: {
    combineReducers: reducers => {
      return combineReducers({
        ...reducers,
        site: undoable(reducers.site), // only this model should be undoable.
      })
    },
  },
})
```

Squeegy:
Yep that was exactly what I needed. Thanks for telling me to look there!

vimcaw:
> That was a good hint! So I think I'm on the right track with something like this?
> 
> ```js
> import { init } from '@rematch/core'
> import { combineReducers } from 'redux'
> import undoable from 'redux-undo'
> import models from './models'
> 
> export const store: Store = init({
>   models,
>   redux: {
>     combineReducers: reducers => {
>       return combineReducers({
>         ...reducers,
>         site: undoable(reducers.site), // only this model should be undoable.
>       })
>     },
>   },
> })
> ```

I tried this config in rematch 1.2.0 and redux-undo 1.0.0-beta9-9-7, but it doesn't work and cause many problems, such as some component doesn't rerender after the used model updated.

vimcaw:
even I just set `combineReducers` to `reducers => combineReducers(reducers)`, nothing else, it still causes many problems, such as some component doesn't rerender after the model they used updated.

```js
import { init } from '@rematch/core'
import { combineReducers } from 'redux'
import immerPlugin from '@rematch/immer'
import createLoadingPlugin from '@rematch/loading'
import models from './models'

const loading = createLoadingPlugin({});
const immer = immerPlugin();

export const store: Store = init({
  models,
  redux: {
    combineReducers: reducers => combineReducers(reducers)
  },
  plugins: [immer, loading]
})
```

vimcaw:
[...]

PR #572, opened 2018-11-11
[Proposal][WIP]Add a declarative redux adaptor

d3dc:
## Proposal
This PR adds a new plugin `@rematch/port`. 

`port` is a plugin that exposes several declarative options to use tested, hopefully type-safe, transformations of the store's config. You call it with a flat object and the plugin does all the config merging.

## Motivation
Several plugins are untested. Things like `react-navigation` are hard to test because of their different build stack.

By putting the rematch wiring in its own wiring, unit tests only ~~need to~~ _should_  be written for the binding code in any complex plugins while most redux "ports" have no additional surface area.

<sup><sup>We also have a recipe folder that I guess you just copy-and-paste out of.</sup></sup>

### Uses
- port **any** existing redux library inline
- package and encapsulate legacy code
- makes heavyweight plugins trivial or unnecessary
- enables "presets" to include redux and legacy code declaratively

### Example

The "Where's my flavor of router?":
```js
const store = init({
  plugins: [
    ...
    port({
      modelName: 'router',
      reducer: connectRouter(history),
      middleware: routerMiddleware(history),
      actionCreators: routerNavigationActions
    })
  ]
}
```

#509
```js
function rematchUndoPlugin({ undoable = [] }) {
  const actionCreators = undoable.reduce(
      (actions, key) => {
        actions[key] = () => undo(key)
        return actions
      },
      {}
    )

  const mapReducers = reducers =>
      undoable.reduce(
        (nextReducers, key) => {
          nextReducers[key] = undoableReducer(reducers[key])
          return nextReducers
        },
        { ...reducers }
      )

  return port({
    modelName: 'undo',
    actionCreators,
    mapReducers
  })
}
```

### API
```ts
interface PortConfig {
	modelName?: string
	reducer?: Reducer
	initialState?: any
	middleware?: Middleware
	actionCreators?: ActionCreatorsMapObject
	combineReducers?: typeof combineReducers
	mapReducers?: (reducers: ReducersMapObject) => ReducersMapObject
	mapRootReducer?: (rootReducer: Reducer) => Reducer
	expose?: {} | ((store: RematchStore) => {})
}
```

## Spaghetti Dispenser
So ports reduce the boilerplate for every library that could be brought to the table. It would make sense to have a way to quickly publish a newly configured "port"
```sh
$ npm run port
```

This creates a new "port recipe". When published, this can be used as:
```js
import recipe from '@rematch/port/recipe/myRecipe'
```

I currently have this so rollup will simply package a `cjs` and `esm` build of each ported recipe and include them in the npm package as separate scripts. babel et. al. tend to use individual packages, and my idea may be a mistake.

### Why my idea may be a mistake
- Dependencies are "implicit" - not listed in the package.json
- It'd be neat to run tests for each dependency
  - How do we include the dependencies of each recipe?

I guess one solution would just be adding them in the plugin root package before or after you `npm run port`. They would be peer and dev depencies for testing. _Importing a recipe would have to rely on implicits or produce a lot of peer log spam_

###  What I don't know

Are separate packages a maintainability nightmare?

## Counter-examples
The "Where's my flavor of router?":
```ts
const plugin = {
  config: {
    models: {
      router: {
        state: null,
        baseReducer: connectRouter(history),
        effects: dispatch =>
          bindActionCreators(routerNavigationActions, dispatch)
      }
    }
  },
  middleware: routerMiddleware(history)
}

const store = init({
  plugins: [plugin]
})
```


d3dc:
# Notable Problem
If I rename `port.test.js` to `port.test.ts` I run into type errors because:
- bad dispatcher typings (I guess this is my next fix)
- no extracted typings for extra plugin options

## Notable Good
using `ts-jest` provides a very informative testing experience

ShMcK:
Clever idea, & well documented. I can tell @d3dc you're the type of person who can automate himself out of a job. To be clear, that's a big compliment 🥇 

What I do know? Yes, separate packages can become a nightmare with breaking changes. In a stable ecosystem, separate packages seem to work fine.

Happy to merge when you feel this PR is ready.

d3dc:
@ShMcK I'm curious on your thoughts whether this should just be in some utils package.

There's a lot of repeated `mapValues` etc. boilerplate in the plugins. I'd wonder if these should just be batteries included in core...

One advantage to a separate package is it decouples things to a stable(?) api.

Issue #565, opened 2018-11-01
Conflicting LICENSE (MIT LICENSE, ISC SPDX)


dcousens:
In https://github.com/rematch/rematch/pull/24 it is clear you (@blairbodnar, @ShMcK) chose MIT as your license,  and your LICENSE file is that.

But your SPDX identifier on `npm` is released as `ISC`.

dcousens:
Ref: https://github.com/rematch/rematch/blob/b3e809836696dd0b9948a6ddfe7aef268ec5e8bd/package.json#L16

dcousens:
The same is for each of your plugins.
Are they ISC or MIT?

millette:
ISC is npm's (strange) default. A PR would probably be welcome :-)

PR #576, opened 2018-11-20
license is MIT, not ISC

dcousens:
Resolves https://github.com/rematch/rematch/issues/565#issuecomment-440131754


ShMcK:
Thanks for the quick fix :)

dcousens:
The plugin SPDX identifiers still need to be fixed

PR #579, opened 2018-11-24
Update plugin licenses to "MIT"

ShMcK:
Resolves follow up issue in #576.

Will also require minor updates to be published.


dcousens:
You aren't distributing the MIT license with your npm releases - I think you should probably add [and publish] https://github.com/rematch/rematch/blob/master/LICENSE to each plugin too.

Annoying,  but

> The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.

PR #605, opened 2019-01-22
Cloning state

ityuany:
Cloning state prevents users from modifying state directly in effects

### DIFF
--- a/src/plugins/effects.ts
+++ b/src/plugins/effects.ts
@@ -54,9 +54,13 @@ const effectsPlugin: R.Plugin = {
 			// async/await acts as promise middleware
 			if (action.type in this.effects) {
 				await next(action)
+
+				const deepClone = (targetObject: any): any =>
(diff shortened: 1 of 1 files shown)


d3dc:
Thanks for contributing!

I'm wondering if a shallow clone actually prevents this.

```js
state.model.value = 42
```

CN-YUANYU:
@d3dc     sorry , It has been revised to deepClone

antongolub:
@CN-YUANYU @d3dc, stringify-parse would not be the best choice for circular refs.
```
const foo = {}; foo.foo = foo; JSON.stringify(foo)
```

d3dc:
Rather than adding a lot of potential overhead to effects, it might be better to take a step back.

I think there is a concern that its not obvious you should not modify state in effects - or maybe its just easy to do. If we modify the clone, our changes just "disappear".

Perhaps a shallow comparison at the end of running the model effect and a development warning would be a better DX.

PR #627, opened 2019-03-29
fix: fix type lost for reducer and effects

Pennnn:
The current d.ts cannot export correct payload type of reducers or effects. 
So i export the type of payload reducers or effects.(But i don't know why have to make a  statement for `boolean`


semoal:
For my opinion, have no sense to specify the payload it's a boolean when it's not.
Could you provide some more information about this? Thx

JiangWeixian:
i need this pr....

ShMcK:
The boolean type in this PR seems to be taking preference over other types. See the error in the build:

```
Error message: "Semantic: /home/travis/build/rematch/rematch/examples/ts/count/src/App.tsx (14, 51) Argument of type '1' is not assignable to parameter of type 'boolean'.
    
    Semantic: /home/travis/build/rematch/rematch/examples/ts/count/src/App.tsx (15, 61) Argument of type '1' is not assignable to parameter of type 'boolean'."
```

PR #630, opened 2019-04-10
Keep isEffect tag

kostychev:
Fixes isEffect tag loss in loading plugin.

### DIFF
--- a/plugins/loading-immutable/src/index.ts
+++ b/plugins/loading-immutable/src/index.ts
@@ -138,6 +138,8 @@ export default (config: LoadingConfig = {}): any => {
 					}
 				}
 
+				effectWrapper.isEffect = true
+
 				// replace existing effect with new wrapper
 				this.dispatch[name][action] = effectWrapper
 			})
--- a/plugins/loading/src/index.ts
+++ b/plugins/loading/src/index.ts
@@ -141,6 +141,8 @@ export default (config: LoadingConfig = {}): Plugin => {
 					}
 				}
 
+				effectWrapper.isEffect = true
+
 				// replace existing effect with new wrapper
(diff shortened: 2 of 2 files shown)


semoal:
Hi, @kostychev could you provide some information about this pull request? 
What's the purpose? Is it a feature or a bug fix?

Thanks in advance.

d3dc:
@semoal the `isEffect` tag is added by the core Effects plugin and used to watch for when to run side-effects. I haven't used the loading plugin to know the bug.

semoal:
> @semoal the `isEffect` tag is added by the core Effects plugin and used to watch for when to run side-effects. I haven't used the loading plugin to know the bug.

I've used loading plugin in some project, but didn't find any issue related to this. That's why am i asking ^^

byzyk:
I have stumbled upon the same issue recently.

I was working on a custom plugin where I had to rely on `isEffect` flag and there was an edge case: if the plugin is injected *after* the loading plugin like this `plugins: [loadingPlugin, myCustomPlugin]` - then `isEffect` flag in `myCustomPlugin` gets lost.

Came here to submit exactly the same PR but then found this one, I guess this should be merged in 👍🏽

matepaiva:
"Came here to submit exactly the same PR but then found this one, I guess this should be merged in ". 

Please merge as soon as possible. It was very hard to identify this bug.

semoal:
> "Came here to submit exactly the same PR but then found this one, I guess this should be merged in ".
> 
> Please merge as soon as possible. It was very hard to identify this bug.

it's already merged and probably already release on rematch 1.2.0

matepaiva:
Oh, really. My fault, then. I will update my packages.

PR #634, opened 2019-05-26
(fix): fix some vulnerabilities by npm

semoal:
Running npm install was arguing about some high vulnerabilities on some packages, was an easy fix with audit.

I passed tests and did some comparison on my actual projects and looks nothing has been broken.

I don't like a project like this, that is amazing npm argues about vulnerabilities (when they're not relevant but anyways it's cool to be it updated)


semoal:
I close this PR because my fork was outdated. BRB

PR #644, opened 2019-07-05
[WIP] Use lerna

d3dc:
This PR moves the entire repo over to a familiar "mono-repo" style. IMO, this makes it easier for people to fork and make changes, as well as experiment with private packages.

### Things that have changed:
- Core is just another package and doesn't build and test the plugins
- I made some assumptions about rollup and abstracted the repeated config to a private package
- Every package.json files got cleaned up
- the script that cleans _and_ build is moved to "prepare" - **this is a lifecycle script**
- I removed any leftover yarn lint.

### Things that are blocking:
The **prepare lifecycle** happens post-install so that packages can be bootstrapped within the mono-repo. The idea then is that `master` always builds fully. 

- [ ] select plugin tests fail with a typescript generics error. bails before building.
- [ ] Contributors file/list in package.json
- [ ] update package homepages etc. for new file tree


.travis.yml
```diff
@@ -11,13 +11,13 @@ jobs:
...
+        - npm run install
         - npm run build
       script: npm run test
       after_script:
         - npm run size
         - npm install coveralls
         - coveralls < coverage/lcov.info          
     - stage: typings
-      before_install: npm run clean --prefix examples/ts/count
+      before_install: npm run clean
```
d3dc: I'm not quite sure what needs to be done in CI

packages/rollup-config/index.js
```diff
@@ -0,0 +1,77 @@
...
+	const name = 'rematch.' + namespace || pkgNamespace
+
+	const rollupBuild = {
+		input,
+		plugins: env === 'production' ? productionPlugins : plugins,
+		external: pkgDependencies,
+		output: [],
+	}
+
+	if (iife) {
```
d3dc: This is a big change. UMD is just more weight if this is only our "Browser" file, since UMD includes cjs.

semoal:
Nice Sam! I’ll take a look of this at night.

Was a change that I spoke with Shawn some weeks ago and couldn’t start yet. If you need help with something create a task list and I’m glad to help.

I’m updating all the docs to new Gitbook UI too. Let’s give a push to this amazing project

d3dc:
One thing that should be fixed here is contributors. Each package should correctly identify who contributed!

We can use an AUTHORS file:

> If there is an AUTHORS file in the root of your package, npm will treat each line as a Name <email> (url) format, where email and url are optional. Lines which start with a # or are blank, will be ignored.

or the package.json `author` and `contributors` field like I did so far. (the old `authors` field isn't in spec?)

semoal:
@d3dc Hi Sam, this pull request is really interesting for the maintance of Rematch, i can help you with polishing last things before merge.

Could you fix conflicts then we check out what's left?

d3dc:
I'll try to get to this this weekend. This is definitely a crunch week for me.  :)

Issue #601, opened 2019-01-13
[TypeScript]: Dispatch doesn't infer reducer or effects functions


WestBrian:
I have followed the TypeScript example exactly but my `dispatch` object isn't inferring my reducer or effects functions. Oddly enough, it was working when I just had effects and no reducer functions. Here is the code:

`store.ts`
```typescript
import { init, RematchRootState } from "@rematch/core";
import * as models from "./models";

const store = init({ models });

export type Store = typeof store;
export type Dispatch = typeof store.dispatch;
export type IRootState = RematchRootState<typeof models>;

export default store;
```

`auth.ts`
```typescript
import { createModel } from "@rematch/core";

interface State {
  user: any;
  errors?: AuthErrors;
}

const initialState: State = {
  user: {}
};

export const auth = createModel({
  state: initialState,
  reducers: {
    signUpError: (state: State, payload: ErrorPayload): State => { ... }
  },
  effects: {
    async signUp(payload: SignUpPayload) { ... }
  }
});
```

`Component.tsx` (I had to cast the dispatch object because I was getting an error that Redux's dispatch object was not equal to Rematch's dispatch object)
```typescript
import { IRootState, Dispatch } from "../../state/store";

...

const mapDispatch: MapDispatchToProps<IDispatchProps, IOwnProps> = (
  dispatch: any
) => {
  const d = dispatch as Dispatch;

  return {
    signUp: (payload: SignUpPayload) => d.auth.signUp(payload)
  };
};
```

The error occurs at the `d.auth.signUp(payload)` line saying:
```
[ts] Argument of type 'SignUpPayload' is not assignable to parameter of type 'boolean'. [2345]
```

It doesn't seem to infer anything from the auth object. The `signUp` function has some absurdly long type and is saying that it's payload is a boolean which is incorrect. As mentioned before, if I remove the `reducers` object in the `ModelConfig` then it infers the auth model correctly.

danawoodman:
Similar issue here. Getting:

```
[ts] Property 'setLanguage' is missing in type '{}' but required in type 'Readonly<Pick<Props, "setLanguage">>'
```

My model looks very similar to the above and I have an async effect called `setLanguage` which is used by the `connect`ed component. When I consume the component, I get the above error.

Have followed the Typescript guide to a T and triple checked everything

canyavall:
Hi WestBrian, I had the same issue than you and my problem was that createModel was wrapping an object with properties and not the constant of the object.
Can you try this?

[...]

Issue #638, opened 2019-05-29
ts count example type error


JiangWeixian:
![image](https://user-images.githubusercontent.com/6839576/58525629-07433780-81ff-11e9-998a-0547cf76a384.png)
![image](https://user-images.githubusercontent.com/6839576/58525658-23df6f80-81ff-11e9-9cb5-7b840f5fa90c.png)

PR #649, opened 2019-07-11
feat add type-safe ts example working in strict mode

hardfist:
the current example is not working in strict mode, and it's not easy to fix the typing problem, so I create a example working in strict mode for people want to use rematch in strict mode

fix https://github.com/rematch/rematch/issues/601 https://github.com/rematch/rematch/issues/638

type check   for reducer && effect
image


examples/ts/strict-count/src/models/util.d.ts
```diff
@@ -0,0 +1,47 @@
...
+  state: infer S,
+  ...payload: infer P
+) => any ? (...payload: P) => S: () => void
+
+
+type Effect2connect<E extends Function> = E extends (
+  payload: infer P,
+  ...args: any
+) => any
+  ? (payload: P) => Promise<void>
```
xcatliu: `  ? (payload: P) => Promise<void>`
should be
`  ? (payload?: P) => Promise<void>`

Because sometimes an effect may only use the second argument `rootState`:

```
effects: (dispatch: RootDispatch) => ({
    async foo(_payload: any, rootState: RootState) {
        doSomethingWith(rootState);
        dispatch.dolphins.setSomething();
    }
});
```

In this case when we call `foo` in an react component, we'll not pass any arguments:

```ts
foo();
```

So the payload argument should be optional

hardfist: I fix some problems, I think it's ok now

xcatliu: This issue still exist, I created a demo in https://github.com/xcatliu/rematch/tree/hardfist/yj-dev/examples/ts/strict-count
You can clone it and repro this issue

image

xcatliu: Add `?` in line 45 will solve this issue

hardfist: oh I know what you mean now, I don't think   `doubleAsync()` should pass the type check, you can just call `doubleAsync(undefined) `, consider the following two methods
```
async doubleAsync1(_payload: any, rootState: RootState) {
			await delay(500)
			dispatch.sharks.increment(rootState.sharks)
}
```
```
async doubleAsync2(_payload: any, rootState: RootState) {
                        console.log('payload:', payload);
			await delay(500)
			dispatch.sharks.increment(rootState.sharks)
}
```
call `doubleAsync1()` ignoring first parameters may seems ok ,but it's not ok to call `doubleAsync2()`, so allow call `doubelAsync()` may cause problem for doubleAsync2

semoal:
Ci is failing could you take a  look?

hardfist:
> Ci is failing could you take a look?

image
It seems caused by latest typescript@3.6, and it woks fine in typescript@3.5, so i specify the version for ci temporarily

hardfist:
another promblem caused by redux's  bug see https://github.com/reduxjs/redux/issues/3466 , after  upgrade to 4.0.4, it works fine

hardfist:
@semoal can i merge this ?

hardfist:
It seems that count example is strict now, I will close this

hardfist:
ooh, the current two example still exists problem

ShMcK:
Sounds good, more examples seems helpful. 

I do find the `.bind(null, value)` a kind of code smell in this case, otherwise I'd recommend replacing the current version.

Issue #656, opened 2019-07-27
fix: better typescript experience


fnknzzz:
Our team has been migrating into rematch for about two month,  feels excellent. But finding some typing confusing. 

- Now in the repo running `npm run test:ts` would not passed. 

- When dispatching, it seems lost typing constrains. For example, now in `example/ts/count/src/App.tsx`, in  `mapDispatch`, you can dispatch whatever you want and not receive any errors. Like dispatching an string, but number is what we want.
![image](https://user-images.githubusercontent.com/22239935/61991905-7a471e00-b089-11e9-844b-56e95c3025f2.png)

- Lost model fileds constrains like this in same file:
![image](https://user-images.githubusercontent.com/22239935/61991923-d3af4d00-b089-11e9-822a-3a7a23448cc9.png)

ShMcK:
Should be resolved with @rematch/core v1.2 in #657.

PR #657, opened 2019-07-29
fix: improved type checking & typescript example

fnknzzz:
1. More accurate type

1. Fix latest redux typing check error in typescript example

1. Add more type constrain in typescript example

related for issue #656

### DIFF
--- a/examples/ts/count/src/store.ts
+++ b/examples/ts/count/src/store.ts
@@ -1,10 +1,10 @@
-import { init, RematchRootState } from '@rematch/core'
-import * as models from './models'
+import { init, RematchRootState, RematchDispatch } from '@rematch/core'
+import models from './models'
 
 export const store = init({
 	models,
 })
 
 export type Store = typeof store
-export type Dispatch = typeof store.dispatch
+export type Dispatch = RematchDispatch<typeof models>
 export type iRootState = RematchRootState<typeof models>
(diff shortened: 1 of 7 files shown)


semoal:
Thanks for your work mate.

Probably this is duplicated with:
https://github.com/rematch/rematch/pull/627
https://github.com/rematch/rematch/pull/647

We should see which one we need.

fnknzzz:
> Thanks for your work mate.
> 
> Probably this is duplicated with:
> #627
> #647
> 
> We should see which one we need.

OK, we can test more tyepscript case

mgambati:
Any progress on review?

examples/ts/count/src/store.ts
```diff
@@ -1,10 +1,10 @@
...
+import { init, RematchRootState, RematchDispatch } from '@rematch/core'
+import models from './models'
 
 export const store = init({
 	models,
 })
 
 export type Store = typeof store
-export type Dispatch = typeof store.dispatch
+export type Dispatch = RematchDispatch<typeof models>
```
semoal: Is it required to create the dispatch method like this?

fnknzzz: No. This is  used in Model defination like `effects: (dispatch: Dispatch) => {}` to avoid recursive check.

ShMcK:
Looks good, sorry haven't been keeping up with PR requests. 

I'll release changes in a batch as 1.2.0.

PR #658, opened 2019-08-02
use ModelEffects<S> instead of ModelEffects<any>

xcatliu:


semoal:
Thanks for your pull request.
Could you explain the reason of this pull request?

PR #659, opened 2019-08-06
Docs (api.md) effect with same name as reducer

hudecsamuel:
## Changes
- Added section to `api.md` docs that explain the order in which effects and reducers with the same name are executed

Behavior from [src/plugins/effects.ts](https://github.com/rematch/rematch/blob/master/src/plugins/effects.ts#L56)


semoal:
@hudecsamuel check conflicts mate, we’ll merge it

hudecsamuel:
@semoal Resolved 🙂

PR #674, opened 2019-09-16
fix global variable name for reselect

ex37:
The umd build for reselect library exposes the lib with capital letter, i.e. `Reselect`, which means current UMD build is broken.


ShMcK:
Thanks @ex37 - I'll publish it shortly

Issue #604, opened 2019-01-22
Object.entries is not compiled in cjs version of @rematch/immer


hardfist:
It seems that @rematch/immer's cjs version is not es5 compatible, since it doesn't compile Object.entries, unfortunately Object.entries is buggy in some chrome versions which broke application.
https://bugs.chromium.org/p/chromium/issues/detail?id=836145
https://stackoverflow.com/questions/49385066/the-strange-behavior-of-object-entries

semoal:
@hardfist is this issue still open or we can close?

hardfist:
> @hardfist is this issue still open or we can close?

still open

hardfist:
1.1.0 version， `rematch-immer.cjs.js ` compiled code contains `object.entries`
![image](https://user-images.githubusercontent.com/8898718/58477198-93187d80-8185-11e9-85a3-240d3ce2da20.png)

PR #680, opened 2019-09-22
fix immer plugin compatible problem

hardfist:


hardfist:
fixs https://github.com/rematch/rematch/issues/604

ShMcK:
Published as `@rematch/immer@1.1.2`

PR #681, opened 2019-09-22
fix the getting started link

madisvain:


ShMcK:
Thanks for the fix :)

PR #686, opened 2019-10-02
Modify the files field in package.json

jaywcjlove:


jaywcjlove:
@ShMcK The type file was not uploaded.

ShMcK:
published as @rematch/loading@1.2.1

PR #690, opened 2019-10-16
Fix escape all string

pedroprado010:
previous escapeRegex would only replace the first occurrence of '*'.


ShMcK:
Thanks :). Will publish later today

PR #711, opened 2019-11-05
Update devtools.md for remote-redux-devtools

dayze:


ShMcK:
Thanks for looking out for future folks with the same problem :)

Issue #697, opened 2019-10-28
Trouble while using enum as payload with Typescript


BlueSky1997AL:
**Describe the bug**
In `RematchDispatcher<P = void, M = void>`, when `P` is enum type, get unexpected dispatch function type. The `extends` operator in type `RematchDispatcher` seems to expand the enum type passed by the generic `P`. This behavior block me from using enum type as payloads.

**To Reproduce**
I declare a model named `test`, payload of its dispatcher `setNewTest` is an enum: `TestEnum`
```ts
enum TestEnum {
    A,
    B,
    C
}

export type TestState = TestEnum;

const model = {
    state: TestEnum.A,
    reducers: {
        setNewTest: (_: TestState, payload: TestEnum) => payload
    }
};

export const test: typeof model = createModel(model);
```
When I use this dispatcher in my code:
```ts
import * as React from 'react';
import { connect } from 'react-redux';

import { iRootState, Dispatch } from './store';

const mapState = (state: iRootState) => ({
    test: state.test
});
const mapDispatch = (dispatch: Dispatch) => ({
    setNewTest: dispatch.test.setNewTest
});

type connectedProps = ReturnType<typeof mapState> & ReturnType<typeof mapDispatch>;
type Props = connectedProps;

class TestComponent extends React.Component<Props> {
    render () {
        return (
            <div style={{ width: 120 }}>
                <h1>{this.props.test}</h1>
                <button onClick={this.props.setNewTest()}>+1</button>
            </div>
        );
    }
}

export default connect(mapState, mapDispatch)(TestComponent);
```
Type of function `this.props.setNewTest` become `(action: Action<TestEnum, void>): Dispatch<Action<TestEnum, void>>` (no overload)

**Expected behavior**
Type of function `this.props.setNewTest` to be `(payload: TestEnum): Action<TestEnum, void> & (action: Action<TestEnum, void>): Dispatch<Action<TestEnum, void>>` so that I can use enum as payload directly

**Screenshots**
![image](https://user-images.githubusercontent.com/28924863/67659786-2dc8f280-f998-11e9-8ebf-dc68f24439df.png)

**Desktop (please complete the following information):**
 - OS: macOS 10.14.6
 - Browser: Chrome 77.0.3865.120
 - Version: @rematch/core@1.2.0

BlueSky1997AL:
See [distributive conditional types](http://www.typescriptlang.org/docs/handbook/advanced-types.html#distributive-conditional-types) for more information

PR #714, opened 2019-11-14
fix: fix type error while passing enum type to `RematchDispatcher` generics

BlueSky1997AL:
#697


ShMcK:
Will release later today, thanks :)

777PolarFox777:
I'm sorry, but you said "later today" about three days ago and still no new release. When will you release new version with this fix?

ShMcK:
I forgot the golden rule: never give deadlines

Thanks for the reminder! It should be published now.

PR #718, opened 2019-12-01
fix: change browser field entry to esm file

hardfist:
It seems that browser field is designed to work with bundler like rollup|webpack, which should not point to umd file。 And it will makes problem when we use webpack to uglify an uglified files. In my app it works fine during development mode , but crash in production because webpack uglify rematch's umd files, which changes js semantics , event though it's should be terser plugin bug, but It dosn't make sense webpack use umd file to bundle app, which unfortunately when webpack's target is browser, `browser` overrides `module` field, which cause problem。
### reference 
* https://github.com/defunctzombie/package-browser-field-spec 
* https://webpack.js.org/configuration/resolve/#resolvemainfields
* https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/264


semoal:
Probably the problem is in plugins packsge.json tío

Can you take a look ?

hardfist:
> Probably the problem is also in plugins packsge.json
> 
> Can you take a look ?

ok, i update all the browser field

semoal:
> > Probably the problem is also in plugins packsge.json
> > Can you take a look ?
> 
> ok, i update all the browser field

I suppose, all the plugins contains esm build on rollup right?
LGTM

hardfist:
> Right now, CI is failing cuz UMD build is not getting compiled, take a look mate

my bad, rollup reads browser fields, I think `unpkg` filed is more appropriate here, just like vue https://github.com/vuejs/vue/blob/dev/package.json#L7

hardfist:
I have to fix all the plugin again , and it seems that some plugin event doesn't have esm output

hardfist:
since the v2 refactor, I will drop this

PR #719, opened 2019-12-02
Update Reactotron/Rematch Integration

Oba-One:
Reactotron had a breaking change removing `createStore` in favor of `createEnhancer()` which now has to be defined on the `enhancers` array instead of the `createStore` function. For typescript with strict typing a null assertion is needed since the type can return undefined.


semoal:
Big thanks Afolabi for the correction!

PR #731, opened 2019-12-27
Add LoadingState type.

mathieutu:
Hi, 
Rematch newcomer here!

I'm using Typescript, and would like to use the loading plugin.
However, because the loading model is added with the plugin api, the AppState (`RematchRootState<typeof models>`) type doesn't know about the loading state.

I've added it in the plugin, to be able to import it and use it directly: 

`export type AppState = RematchRootState<typeof models> & LoadingState<typeof models>`

I use the type on my side, but share it with the community, as I think it can interest.

Thanks for this cool package,
Matt'.

### DIFF
--- a/plugins/loading/src/index.ts
+++ b/plugins/loading/src/index.ts
@@ -7,6 +7,18 @@ export interface LoadingConfig {
 	asNumber?: boolean
 }
 
+export interface LoadingState<M extends Models> {
+	loading: {
+		global: boolean,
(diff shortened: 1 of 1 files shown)


ShMcK:
Typings look great, thanks :)

mathieutu:
Thanks for the fast merging. Actually this is a first version, which doesn't handle any eventual default configuration overwriting. I'll let someone interested upgrades it if needed.

Issue #594, opened 2018-12-26
[discuss] the way of use dispatch  impact the loading plugin


zhangwilling:
here are ways of dispatch using  in effects: 

  1. dispatch({ type: 'count/incrementAsync', payload: 1 })

  2. dispatch.count.incrementAsync(1)

the first will not emit loading, but second is ok.

a little bit confusing.

jackwilsdon:
Duplicate of #588

Issue #588, opened 2018-12-04
Dispatch with object type not trigger loading/show and loading/hide


tsl:
I found a issue when I using plugins: [loadingPlugin], the action loading/show not trigger, when I dispatch type object like dispatch({ type: 'model/actionName' }), it will trigger when I dispatch.model.actionName.

Case dispatch with object type:

![case-1 1](https://user-images.githubusercontent.com/14858308/49436567-08fde680-f7ec-11e8-80d0-795b6eaa4943.jpg)
![case-1 2](https://user-images.githubusercontent.com/14858308/49436584-10bd8b00-f7ec-11e8-8a72-1f0824581389.jpg)

Case dispatch with mapDispatch:

![case-2 1](https://user-images.githubusercontent.com/14858308/49436618-2337c480-f7ec-11e8-8eb0-8057006c03bb.jpg)
![case-2 2](https://user-images.githubusercontent.com/14858308/49436627-2632b500-f7ec-11e8-8c97-aa1052df9374.jpg)

nganhkhoa:
I think we should fix this, it is more Redux to write `dispatch(ACTION)`. I don't understand why `distach(ACTION)` fails to set loading, because `dispatch.models.effects` is just an alias to `dispatch(ACTION)`.

[createDispatcher](https://github.com/rematch/rematch/blob/ddd44c1e97b56f8843f55e4ea4c4511a6a8358fe/src/plugins/dispatch.ts#L36)

[effects/onModels](https://github.com/rematch/rematch/blob/ddd44c1e97b56f8843f55e4ea4c4511a6a8358fe/src/plugins/effects.ts#L42)

```js
this.dispatch[model.name][effectName] =
	this.createDispatcher.apply(
		this,
		[model.name, effectName]
	)
```

PR #745, opened 2020-03-12
Make different dispatch calling effect equivalently

blacktail:
Currently, the following two kinds of dispatch call is not equivalent:

```
dispatch.modelName.actionName(payload, meta)
```

vs
```
dispatch({
  type: 'modelName/actionName',
  payload,
  meta,
}
``` 

And this will introduce some issues into other plugins , for example the loading plugin: #588 , #594.


AdamCloud8:
Is there any plan on resolving this?

semoal:
> Is there any plan on resolving this?

Absolutely, if it's still an issue, of course. Do you experience the same issue? Could you open an issue and we track the issue.

PR #771, opened 2020-06-23
docs/plugins/persist.md line[63] should Change bracket to parentheses

little-buddy:
should be:
```
const Root = () => (
	<PersistGate persistor={persistor}>
		<App />
	</PersistGate>
)
```
not:
```
const Root = () => {
	<PersistGate persistor={persistor}>
		<App />
	</PersistGate>
}
```


ShMcK:
Thanks, good catch 👍

PR #788, opened 2020-07-28
fix: incorrect type inference of model dispatchers

tianzhich:
When defining the dispatch param type inside the model, you should pass the generic `RootModel` type like `M`(or `RM`) to `ModelConfig<any, any, M>`. After this, the user-defined model can extend from `ModelConfig<any, any, M>` and type inference would be correct.

### DIFF
--- a/src/typings/index.d.ts
+++ b/src/typings/index.d.ts
@@ -11,7 +11,7 @@ import * as Redux from 'redux'
 export as namespace rematch
 
 export type ExtractRematchStateFromModels<M> = {
-	[modelKey in keyof M]: M[modelKey] extends ModelConfig ? M[modelKey]['state'] : never
+	[modelKey in keyof M]: M[modelKey] extends ModelConfig<any, any, M> ? M[modelKey]['state'] : never
 }
 
 export type RematchRootState<M> = ExtractRematchStateFromModels<M>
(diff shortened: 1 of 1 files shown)


semoal:
This has been fixed on v2.0.0-next branch, thanks for the help mate.

haikyuu:
@tianzhich I'm trying your PR but I don't know how to make it work.

Can you give an example? You can fork this sandbox for instance.

Thank you

tianzhich:
@haikyuu You can try https://codesandbox.io/s/rematch-typescript-8mit8?file=/src/rematch.d.ts. This is the demo code of [using ts with rematch](https://github.com/rematch/rematch/tree/master/examples/ts/hooks). And I override the typings. If you delete it you could see the original error.

The main difference was: 

image

But I don't have the time recently. I will try v2 later.

semoal:
> @haikyuu You can try https://codesandbox.io/s/rematch-typescript-8mit8?file=/src/rematch.d.ts. This is the demo code of [using ts with rematch](https://github.com/rematch/rematch/tree/master/examples/ts/hooks). And I override the typings. If you delete it you could see the original error.
> 
> The main difference was:
> 
> image
> 
> But I don't have the time recently. I will try v2 later.

Feel free to take a look the help-wanted issue on next branch. We need some help on the dispatch inference.
I've tried my best without luck =)

PR #799, opened 2020-08-12
Update typescript section with more instructions

haikyuu:
Add a note for the user to type the initial state and reducers to get full typing benefits.


semoal:
This is not necessary anymore mate, thanks for contribution anyways!

Issue #797, opened 2020-08-11
Typescript extractions broken in next (2) version


haikyuu:
Note I'm using the `2.0.0-next.1` version of rematch.

**Describe the bug**
Duplicate of https://github.com/rematch/rematch/issues/787
This bug has been reported before and a fix was suggested here for instance https://github.com/rematch/rematch/pull/788. But it was closed as already fixed. And it seems it's still an issue.

**Expected behavior**
Dispatch and state types should be extracted correctly.

**Screenshots**

![image](https://user-images.githubusercontent.com/8558836/89933027-5a3bbb80-dc06-11ea-89b6-4b9edd8d6823.png)
![image](https://user-images.githubusercontent.com/8558836/89933053-6293f680-dc06-11ea-9403-2b422b4894df.png)
![image](https://user-images.githubusercontent.com/8558836/89933089-72133f80-dc06-11ea-97c5-a842e7d7b256.png)
![image](https://user-images.githubusercontent.com/8558836/89933276-ba326200-dc06-11ea-80d6-b3a0ba991590.png)

semoal:
Could you provide us a reproduction repo? 

In this way we can debug better the issue

haikyuu:
@semoal Sure. I'll create one

haikyuu:
@semoal here it is. https://codesandbox.io/s/gifted-shape-6oywz?file=/src/models/questions.ts

semoal:
Thanks @haikyuu we'll look asap onto it

semoal:
If you need a workaround i've fixed it haikyuu, we'll look into extracter type of the state, because is the type that is failing.

Changed to this:

```js
  state: {
    questions: [] as Array<BooleanQuestion>,
    amount: 2,
    type: "boolean" as QuestionType
  }, // initial state
```

```
// state.d.ts
type QuestionType = "boolean" | "multiple" | "mixed";
```

haikyuu:
@semoal Thanks a lot @semoal

haikyuu:
@semoal 
> we'll look into extracter type of the state, because is the type that is failing.

Note that `Dispatch` types are also failing

haikyuu:
@semoal When declaring the initial state and typing it in the upper scope, typing works just fine. I've updated the codesandbox and it works fine now.

It's probably an issue of type extractions needing types to be explicitly defined rather than inferred. And that's fine IMHO.

Another thing is that the user needs to type reducers and effects parameters to get their typings in other places in the app.

[...]

PR #800, opened 2020-08-16
Next types

semoal:
### Done
- Introduced a more complex example where the state is not just numbers, and the types are failing.
- Refactored createModel to add intellisense on state values, issue related with additional screenshots: https://github.com/rematch/rematch/issues/797

## Todo
- [x] Fix the types for dispatch
- [x] Add a test-suite for this case, to avoid a regression.


tianzhich:
Hi @semoal, I am a newbie with Github PR progress. I made a PR and thought it would be continued from here. But the new one was generated #801

semoal:
> Hi @semoal, I am a newbie with Github PR progress. I made a PR and thought it would be continued from here. But the new one was generated #801

no problem mate, i'll merge all the pr you need to this branch

tianzhich:
Thanks, mate. I fixed the dispatcher's type warning. But there was a new incompatible error with the `dispatch` param in `connect` function: 

```
Type '(dispatch: Dispatch) => { incrementDolphins: ((payload: number) => Action<number>) & { isEffect: false; }; incrementDolphinsAsync: (() => Promise<void>) & { isEffect: true; }; incrementSharks: () => Action<...>; incrementSharksAsync: () => Promise<...>; incrementSharksAsync2: () => { ...; }; }' is not assignable to type 'MapDispatchToPropsFactory<{ incrementDolphins: ((payload: number) => Action<number>) & { isEffect: false; }; incrementDolphinsAsync: (() => Promise<void>) & { isEffect: true; }; incrementSharks: () => Action<...>; incrementSharksAsync: () => Promise<...>; incrementSharksAsync2: () => { ...; }; }, {}>'.
        Types of parameters 'dispatch' and 'dispatch' are incompatible.
          Type 'Dispatch<Action<any>>' is not assignable to type 'RematchDispatch<RootModel>'.
            Type 'Dispatch<Action<any>>' is missing the following properties from type 'ExtractRematchDispatchersFromModels<RootModel>': dolphins, sharks, questions
```

Maybe it was duplicated with https://github.com/rematch/rematch/pull/720#issuecomment-630353830.

You can check the count example for more details.

tianzhich:
In the previous codebase, if the dispatch param was defined as `Dispatch`. Typescript will think they are incompatible.

```typescript
export const dolphins = {
	state: 0,
	reducers: {
		increment: (state: DolphinsState, payload: number) => state + payload,
	},
	effects: (dispatch: Dispatch) => ({
		async incrementAsync(): Promise<void> {
			await delay(500)
			dispatch.dolphins.increment(1)
		},
	}),
}
```

Because the [function parameters bivariant](https://github.com/Microsoft/TypeScript/wiki/FAQ#why-are-function-parameters-bivariant). The `Dispatch` that is `Rematch<RootModel>`, should extends `Rematch<TModels>` properly, and vise versa.

I pass the generic param `TModels` using `RootModel` across the type inference progress. So they are compatible and the types for dispatcher could be correct.

semoal:
> In the previous codebase, if the dispatch param was defined as `Dispatch`. Typescript will think they are incompatible.
> 
> ```ts
> export const dolphins = {
> 	state: 0,
> 	reducers: {
> 		increment: (state: DolphinsState, payload: number) => state + payload,
> 	},
> 	effects: (dispatch: Dispatch) => ({
> 		async incrementAsync(): Promise<void> {
> 			await delay(500)
> 			dispatch.dolphins.increment(1)
> 		},
> 	}),
> }
> ```
> 
> Because the [function parameters bivariant](https://github.com/Microsoft/TypeScript/wiki/FAQ#why-are-function-parameters-bivariant). The `Dispatch` that is `Rematch<RootModel>`, should extends `Rematch<TModels>` properly, and vise versa.
> 
> I pass the generic param `TModels` using `RootModel` across the type inference progress. So they are compatible and the types for dispatcher would be correct.

The thing I try to also focus is the requeriment of puting the as on the state, 

```typescript
	state: {
		questions: [] as number[],
		amount: 2,
		type: 'boolean' as QuestionType,
	},
```

Removing the extends from the createModel function gets resolved but dispatch fails :/ , I don't understand well why an extends doesn't infer correctly on the state, when we're passing them all the generics.

tianzhich:
Sorry about that. When creating a model using `createModel`, there are 2 places where we need for state inference, one is the state itself, another is the state param in reducers function.

```typescript
const questions = createModel<QuestionsState>()({
        // place 1, state itself
	state: {
		questions: [],
		amount: 2,
		type: 'boolean',
	},
	reducers: {
		// place 2, state param
		setQuestions(state, payload: Array<number>) {
			return { ...state, amount: 1 }
		},
	}
})
``` 

But if you using the type definition below, you will only get the correct state type inference. So you define the state param type explicitly.

```typescript
export declare const createModel: <State, BaseState = State>() => (
	mo: Model<State, BaseState>
) => Model<State, BaseState>
```

But I think we should get both type inference automatically. So I change to this:

```typescript
export declare const createModel: <TModels>() => <
        State, BaseState = State, 
	M extends Model<State, BaseState, TModels>
>(
	mo: M
) => M
```

image

It will infer correctly on state cause the returned model is the same as the passed model param. So I put `as number[]` requirement on an empty array to avoid default `never[]` inference on it.

But the state param still cant be inferred if we do not define them explicitly. So I pass the State param like below: 

```typesscript
export declare const createModel: <TModels, State, BaseState = State>() => < 
	M extends Model<State, BaseState, TModels>
>(
	mo: M
) => M
```

image

So the `State` generic param is for state param inference and `as number[]` is for the state itself.

I will let them both depend on `State` generic param later.

By the way, I think we can let both state inferred automatically. We can extract reducers type for generic param and do some extra work on the returned model.

tianzhich:
It was broken cause incompatible type defination: 

image

I fixed by using the passed generic `State` to get a consistent and compatible type.

tianzhich:
@semoal  Hi bro, I have rollbacked and recreated a pr #803, can you roll back the current branch and merge from this?

I fixed all the related typing errors in a more clear way. You can check the example, hope that helped.

semoal:
There is still two pending issues:

- [ ] ts-ignore on all the files that must be connected, we should find a way to avoid this. augmentation of redux dispatch or something.
- [ ] complex states with the new createModel is not infered correctly, probably we could do like an optional param on createModel where the state is complex, just pass the type and build the state from that, is that possible?

We're near!!

pjamrozowicz:
> There is still two pending issues:
> 
>     * [ ]  ts-ignore on all the files that must be connected, we should find a way to avoid this. augmentation of redux dispatch or something.
> 
>     * [ ]  complex states with the new createModel is not infered correctly, probably we could do like an optional param on createModel where the state is complex, just pass the type and build the state from that, is that possible?
> 
> 
> We're near!!

Hey guys, I'm a little bit out of the loop right now but just a comment on the augmentation of redux dispatch - I'm not sure if this is possible to do on the library side (maybe it is after the recent changes) but I was trying to figure this out for a looong time without any luck. I'm not sure if you've tried the solution described here - [Step 4: View -> TypeScript](https://rematch.netlify.app/#/quick-start?id=step-4-view), the part where it says:

```
You might need to ‘patch’ Redux types to avoid having issues with libraries such as react-redux...
```

This is the solution to not have the ts-ignore in files with connect method.

tianzhich:
> There is still two pending issues:
> 
> * [ ]  ts-ignore on all the files that must be connected, we should find a way to avoid this. augmentation of redux dispatch or something.
> * [ ]  complex states with the new createModel is not infered correctly, probably we could do like an optional param on createModel where the state is complex, just pass the type and build the state from that, is that possible?
> 
> We're near!!

Well, I'll have a look at the step that @pjamrozowicz mentioned above and try to figure out the redux dispatch. 

But I think we don't need to pass an optional param on `createModel` because we want typescript to do this. When defining complex types like typed arrays or union strings, if we using an empty array, typescript might think it is `never[]`, so we can put `as number[]` to tell typescript it is an array with numbers.

Or by a more elegant way: 

```typescript
type QuestionType = 'boolean' | 'multiple' | 'mixed'
type QuestionsState = {
	questions: number[]
	amount: number
	type: QuestionType
}

const questions = createModel<RootModel>()({
	state: {
		questions: [],
		amount: 2,
		type: 'boolean',
	} as QuestionsState,
        // ...
}
```

Or using a typed initial state outside: 

```typescript
type QuestionType = 'boolean' | 'multiple' | 'mixed'
type QuestionsState = {
	questions: number[]
	amount: number
	type: QuestionType
}

const initState: QuestionsState = {
	questions: [],
	amount: 2,
	type: 'boolean',
}

const questions = createModel<RootModel>()({
	state: initState,
        // ...
}
```

tianzhich:
Glad to tell you guys, I have figured out the incompatibility between RematchDispatch and ReduxDispatch.

Maybe we need to patch the react-redux instead of redux. I extended the `connect` type declaration and added an overload to it: 

```typescript
declare module 'react-redux' {
	interface Connect {
		<
			RM extends Models<RM> = Models,
			State = DefaultRootState,
			TStateProps = {},
			TDispatchProps = {},
			TOwnProps = {}
		>(
			mapStateToProps: MapStateToPropsParam<TStateProps, TOwnProps, State>,
			mapDispatchToProps: MapRematchDispatchToPropsNonObject<
				TDispatchProps,
				TOwnProps,
				RM
			>
		): InferableComponentEnhancerWithProps<
			TStateProps & TDispatchProps,
			TOwnProps
		>
	}

	type MapRematchDispatchToPropsNonObject<
		TDispatchProps,
		TOwnProps,
		RM extends Models<RM> = Models
	> =
		| MapRematchDispatchToPropsFactory<TDispatchProps, TOwnProps, RM>
		| MapRematchDispatchToPropsFunction<TDispatchProps, TOwnProps, RM>

	type MapRematchDispatchToPropsFactory<
		TDispatchProps,
		TOwnProps,
		RM extends Models<RM> = Models
	> = (
		dispatch: RematchDispatch<RM>,
		ownProps: TOwnProps
	) => MapRematchDispatchToPropsFunction<TDispatchProps, TOwnProps, RM>

	type MapRematchDispatchToPropsFunction<
		TDispatchProps,
		TOwnProps,
		RM extends Models<RM> = Models
	> = (dispatch: RematchDispatch<RM>, ownProps: TOwnProps) => TDispatchProps
}
```

It worked!! Typescript will choose this overload to compile rematch.

semoal:
Well guys, that was absolutely amazing, we fixed the issue with typescript in a record time.
@tianzhich brilliant job mate.
@pjamrozowicz for all the previous work, also amazing job dude.

I fixed all the testing suite, and introduced both examples (hooks, and standard state) (before count example was crashing on the test suite, now both are tested, so regression will be easy to check)

Also I've updated the documentation with all the highlight's you found, and added some good additions like all the versions of our current codebase.

thinhtran3588:
Amazing work, guys! 
I've just tried the next version yesterday and also found the problem with Typescript. But after only a day, you guys fixed it!!!
Can't thank you enough for this amazing library 👍

PR #803, opened 2020-08-18
fix: type inference for state and dispatch

tianzhich:


semoal:
why rollback mate?

tianzhich:
I reorganized my thoughts and fix these typing errors by a more clear way.

semoal:
@tianzhich i fixed the conflicts, check that i didn't break anything.
If it's ok i merge

Issue #807, opened 2020-08-23
How to connect Gatsby with rematch?


mossen:
With redux it seems to be easy by following:
https://github.com/gatsbyjs/gatsby/blob/master/examples/using-redux/src/state/createStore.js

but with rematch there is no official document. Can anyone help?

mossen:
@semoal thanks, please share a link here.
Also I do not want to have a wrapper/layout and use that in every single page link. something with wrapWithProvider would be the best I guess.

semoal:
@mossen i've pushed an example with typescript + gatsby + rematch@next.4

![2020-08-26 15 50 58](https://user-images.githubusercontent.com/22656541/91311993-00b1c000-e7b4-11ea-90a1-85588680f9eb.gif)
> Also I do not want to have a wrapper/layout and use that in every single page link. something with wrapWithProvider would be the best I guess.

About this, i'm not an expert with gatsby but i'm 100% sure you can create a wrapper for the page you want.

mossen:
@semoal Thanks for that, that helps. How ever I am looking for a way to not have to wrap every single file with the provider. I will ask this in Gatsby repo as well in case if anyone knows a way.

PR #809, opened 2020-08-27
Fix gatsby example

arthurjdam:
Removes the duplicate Provider. It is already in `src/wrapper.tsx`, which wraps _all_ pages rendered by SSR and in-browser, so the one on `src/pages/index.tsx` may be omitted.

Also adds types for react-redux
And removes the references to src/images, which was preventing the build from starting

Closes last posed question in #807


semoal:
Great Arthur! thanks for your contribution!

PR #810, opened 2020-08-27
Docs Fix URL for "examples" page

arthurjdam:
Currently it refers to: https://github.com/rematch/rematch/examples which returns a 404.

Not sure if this should point to the next or master branch, but it seems most library refer to the newer one of the two.


semoal:
:shipit:

PR #819, opened 2020-09-08
Improved typings

tianzhich:
Improved typings for core, loading, and updated.


packages/core/src/dispatcher.ts
```diff
@@ -46,24 +47,19 @@ const createActionDispatcher = <TModels extends Models>(
  * actions.
  */
 const createDispatcher = <
-	TModels extends Models<TModels>,
-	TModel extends NamedModel<TModels>
+	TModels extends Models<TModels> = Record<string, any>,
```
semoal, line 50: great Record<string, any>

semoal: for typescript 4.0.0 this is the best approach

packages/loading/test/loading-asNumber.test.ts
```diff
@@ -367,7 +367,7 @@ describe('loading asNumbers', () => {
 			reducers: {},
 		}
 		type Models = { count: typeof count2 }
-		type ExtraModels = ExtraModelsFromLoading<Models & ExtraModels>
+		type ExtraModels = ExtraModelsFromLoading<Models>
```
semoal, line 370: we must add this to documentation with a typescript section

packages/updated/test/updated.test.ts
```diff
@@ -32,7 +31,7 @@ describe('updated', () => {
 			},
 		}
 
-		const store = init({
+		const store = init<RootModel, ExtraModelsFromUpdated<RootModel>>({
```
semoal, line 34: same here, we must add this to documentation

semoal (changes requested):

semoal:
If i can this night i can update documentation to match the new extractors on typescript. 
Should we create an example with react and all the plugins?

packages/loading/test/loading-asNumber.test.ts
tianzhich: > we must add this to documentation with a typescript section

yeah, we should, can you initiate one and I'll make some additions if needed.

tianzhich:
> If i can this night i can update documentation to match the new extractors on typescript.
> Should we create an example with react and all the plugins?

Yes, we should, after I improved `@select`. How to create an example in the examples folder? Are there any templates or something? If you can create one, I'll make some improvements.

tianzhich:
BTW, have you already resolved the conflicts about the webpack version? We need to resolve this before `npm start` an example.

packages/core/src/dispatcher.ts
tianzhich: `Record` was great except this: 

```ts
type RematchStoreWithEmptyObject = RematchStore<RootModel & {}>
type RematchStoreWithRecord = RematchStore<RootModel & Record<string, any>>

const store1: RematchStoreWithEmptyObject
const store2: RematchStoreWithRecord

// (property) dolphins: any
// Type of property 'dolphins' circularly references itself in mapped type 'ExtractRematchDispatchersFromModels<RootModel & Record<string, any>>'.ts(2615)
// Type of property 'questions' circularly references itself in mapped type 'ExtractRematchDispatchersFromModels<RootModel & Record<string, any>>'.ts(2615)
// Type of property 'sharks' circularly references itself in mapped type 'ExtractRematchDispatchersFromModels<RootModel & Record<string, any>>'.ts(2615)
store2.dispatch.dolphins.increment

// worked
store1.dispatch.dolphins.increment
```

So I want to change the default generic value for `TExtraModels`, change from this: 

```ts
export declare const init: <
	TModels extends Models<TModels> = Record<string, any>,
	TExtraModels extends Models<TModels> = Record<string, any>
>(
	initConfig?: InitConfig<TModels, TExtraModels> | undefined
) => RematchStore<TModels & TExtraModels>
```

to this

```ts
export declare const init: <
	TModels extends Models<TModels> = Record<string, any>,
	TExtraModels extends Models<TModels> = {}
>(
	initConfig?: InitConfig<TModels, TExtraModels> | undefined
) => RematchStore<TModels & TExtraModels>
```

If our users don't use plugins that contain models, they don't need to pass `RootModel` to init, `RootModel` can be inferred automatically. But if they use Rematch with these plugin models, they need to pass `ExtraModels` info cause it can't be inferred. (Plugin config array contains these model Infos, but I can't find a way to iterate it, collect and combine all of them) And once they pass the `TExtraModel`, they need also pass `RootModel`, otherwise it will be inferred by its opinion value, but not from `initConfig`.

So pass `RootModel` and `ExtraModels` if use plugins that contain models, and don't need to pay attention if don't' use plugins or use plugins that just using hooks like `immer`.

semoal:
> > If i can this night i can update documentation to match the new extractors on typescript.
> > Should we create an example with react and all the plugins?
> 
> Yes, we should, after I improved `@select`. How to create an example in the examples folder? Are there any templates or something? If you can create one, I'll make some improvements.

There isn't any template, just i use create-react-app --template typescript, after that just add a model folder, with store.ts, and start building a little todo app for example.

I'm already fixing versions mismatch, and i'll add a base for the plugins docs with typescript, and a complete example, i'll push to a new branch.
i'll request your review mate

semoal (approved):

semoal:
I'm thinking in shipping all this changes, we fixed all the issues with all the plugins except select, improved docs so probably it's nice to be already tagged.

# Pending
- Fix select plugin as much as we can..
- Change loading plugin behaviour.

PR #820, opened 2020-09-09
Next types docs

semoal:
### Done:
- Added `all-plugins-react-ts` example, (@tianzhich here you have a real example to start with)
- Added an example of how tabs in documentation is done (@tianzhich if you guide me how to integrate the plugins with typescript) Example how is done: https://github.com/rematch/rematch/commit/1bd9169ceed1783bdb976d157e3482e024c865e0
- Updated all dependencies and avoid creact react app webpack mismatch.

### Pending:
- [x] Create a 100% functional example with typescript
- [x] Write plugins docs in typescript


packages/core/package.json
```diff
@@ -27,10 +27,10 @@
     "node": ">=10"
   },
   "peerDependencies": {
-    "redux": "^4.0.5"
+    "redux": "*"
```
tianzhich, line 30: Are we compatible with all versions of redux?

examples/all-plugins-react-ts/src/store.ts
```diff
@@ -0,0 +1,29 @@
...
+import createPersistPlugin from "@rematch/persist"
+import createUpdatedPlugin, { ExtraModelsFromUpdated } from "@rematch/updated"
+import createSelectPlugin from "@rematch/select"
+import storage from 'redux-persist/lib/storage'
+
+type FullModel = ExtraModelsFromLoading<RootModel> & ExtraModelsFromUpdated<RootModel>;
+
+export const store = init<RootModel, FullModel>({
+	models,
+	plugins: [
```
tianzhich, line 15: I will complete plugins after refactored types of `select`.

tianzhich (approved):

packages/core/package.json
semoal: We're not, ill fix this asap

semoal:
I'll give a sprint this week @tianzhich, if we can't close the select part probably we'll recommend if someone is using that plugin use some ts-ignores, but at least we should release a 90% stable branch with all the other plugins.
Lets' finish the v2.0 👍🏻

PR #828, opened 2020-10-14
fix: @rematch/select typescript plugin compatibility

semoal:
### TODO:
- [x] is it a good practice create a createModelWithSelectors<RootModel>()({})? or could we extends native createModel for plugins?

- selectors on model has three props, we must guarantee it doesn't have errors:
	- [X]  selectors: (slice) => ...
	- [ ]  selectors: (slice, createSelector) => ...
	- [ ]  selectors: (slice, createSelector, hasProps) => ...

- [x] store.select is not autocompleted
```jsx
const selection = store.select(models => ({
  // todo: this isn't getting autocompleted
  total: models.cart.total,
}))
```

- [x] second param shouldn't be required, or yes?.. take a look at reselect examples
```jsx
const mapState = (state: RootState) => ({
	settingsState: state.settings,
  loadingState: state.loading,
  playersState: state.players,
  // todo: props shouldn't be required
  ...selection(state, null)
})
```

- [x] Props for class and hooks, should be filled...
```jsx
// todo: we must type the props autocomplete of the selector new values
type Props = StateProps & DispatchProps
```
- [x] Props for class and hooks, should be filled...
```jsx
// todo: we must type the props autocomplete of the selector new values
type Props = StateProps & DispatchProps
```

### DIFF
--- a/packages/select/package.json
+++ b/packages/select/package.json
@@ -24,7 +24,8 @@
     "src"
   ],
   "dependencies": {
-    "reselect": "^4.0.0"
+    "reselect": "^4.0.0",
+    "redux": "^4.0.0"
   },
   "devDependencies": {
     "@rematch/core": "^2.0.0-next.6"
--- a/packages/select/package.json
+++ b/packages/select/package.json
@@ -24,13 +24,14 @@
     "src"
   ],
   "dependencies": {
-    "reselect": "^4.0.0",
-    "redux": "^4.0.0"
+    "reselect": "^4.0.0"
   },
   "devDependencies": {
+    "redux": "^4.0.0",
     "@rematch/core": "^2.0.0-next.6"
   },
   "peerDependencies": {
+    "redux": "^4.0.0",
     "@rematch/core": ">=2.0.0-next.0"
   },
   "husky": {
--- a/packages/select/src/index.ts
+++ b/packages/select/src/index.ts
@@ -1,13 +1,13 @@
 // @ts-nocheck
 import {
 	Models,
-	NamedModel,
 	Plugin,
 	ExtractRematchStateFromModels,
 	Model,
 	ModelReducers,
 	ModelEffects,
 	ModelEffectsCreator,
+	RematchRootState,
 } from '@rematch/core'
 import { Reducer } from 'redux'
 import { createSelector, createStructuredSelector } from 'reselect'
@@ -118,7 +118,7 @@ export const createModelWithSelectors: <RM extends Models<RM>>() => <
 
 const createSelectPlugin = <
 	TModels extends Models<TModels>,
-	TExtraModels extends Models<TModels> = {}
+	TExtraModels extends Models<TModels> = Record<string, any>
 >(
 	config: SelectConfig = {}
 ): Plugin<TModels, TExtraModels> => {
@@ -144,7 +144,7 @@ const createSelectPlugin = <
 	}
 
 	const hasProps = (inner: any) =>
-		function (models: TModels) {
+		function (models: any) {
 			return selectorCreator(
 				(props: any) => props,
 				(props: any) => inner.call(this, models, props)
@@ -161,7 +161,11 @@ const createSelectPlugin = <
 			sliceState,
 			selectorCreator,
 		},
-		onModel(model: NamedModel<TModels>): void {
+		onModel(
+			model: Model<TModels> & {
+				selectors?: ModelSelectorsConfig<RematchRootState<TModels>>
+			}
+		): void {
 			select[model.name] = {}
 
 			const selectorFactories =
--- a/packages/select/src/index.ts
+++ b/packages/select/src/index.ts
@@ -1,7 +1,17 @@
 // @ts-nocheck
-import { Models, NamedModel, Plugin, RematchStore } from '@rematch/core'
+import {
+	Models,
+	NamedModel,
+	Plugin,
+	ExtractRematchStateFromModels,
+	Model,
+	ModelReducers,
+	ModelEffects,
+	ModelEffectsCreator,
+} from '@rematch/core'
+import { Reducer } from 'redux'
 import { createSelector, createStructuredSelector } from 'reselect'
-import { SelectConfig } from './types'
+import { ModelSelectorsConfig, SelectConfig } from './types'
 
 const makeSelect = () => {
 	/**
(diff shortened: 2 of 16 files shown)


packages/select/src/index.ts
```diff
@@ -64,32 +74,80 @@ const validateConfig = (config: SelectConfig): void => {
...
 			throw new Error(
 				`Selector (${model.name}/${selectorName}) must be a function`
 			)
 		}
 	}
 }
 
-const createSelectPlugin = (config: SelectConfig = {}): Plugin => {
+/**
+ * Todo: could we extend types of standard createModel?
```
semoal, line 86: @tianzhich

packages/select/src/types.ts
```diff
@@ -66,15 +76,34 @@ export type RematchSelect<M extends object = Models, RootState = any> = ((
...
+		state: S
+		reducers?: R
+		selectors?: ModelSelectorsConfig<S>
+		baseReducer?: BR
+		effects?: E
 	}
 
-	interface RematchStore<TModels> {
-		select: RematchSelect<TModels, RematchRootState<TModels>>
+	export interface RematchStore<
```
tianzhich, line 101: why copy the same RematchStore in select?

packages/select/src/types.ts
```diff
@@ -66,15 +76,34 @@ export type RematchSelect<M extends object = Models, RootState = any> = ((
...
 	> {
+		name?: string
+		state: TState
 		selectors?: ModelSelectorsConfig<TState>
+		reducers?: ModelReducers<TState>
+		baseReducer?: ReduxReducer<TBaseState>
+		effects?: ModelEffects<TModels> | ModelEffectsCreator<TModels>
+	}
+
+	export interface ModelResult<S, R, BR, E> {
```
tianzhich, line 92: what is the ModelResult used for?

packages/select/src/types.ts
```diff
@@ -66,15 +76,34 @@ export type RematchSelect<M extends object = Models, RootState = any> = ((
 	StoreSelectors<RootState>
 
 declare module '@rematch/core' {
-	interface Model<
+	export interface Model<
```
tianzhich, line 79: I think we don't need to complete the props of Model, these props can be combined when Model definition is used

packages/select/src/index.ts
tianzhich: I was thinking if we create a modelCreator for each plugin for that purpose. How can we use them together?
Could we add a loose index type for `createModel` just like this:

```ts
export const createModel: <
	RM extends Models<RM>
>() => <
	R extends ModelReducers<S>,
	BR extends ReduxReducer<BS>,
	E extends ModelEffects | ModelEffectsCreator<RM>,
	S,
	BS = S
>(mo: {
	name?: string
	state: S
	reducers?: R
	baseReducer?: BR
	effects?: E
	[index: string]: any
}) => {
	name?: string
	state: S
	reducers: R
	baseReducer: BR
	effects: E
        [index: string]: any
} = () => (mo): any => {
	const { reducers = {}, effects = {} } = mo

	return {
		...mo,
		reducers,
		effects,
	}
}
```

So our users can add their own model configs if they needed when creating a new plugin. But cause the type is any, so they need to use `as` to get type completion.

tianzhich: Or we can collect their config into a prop called `configs` so we can pass the type param to get the completion.

packages/select/src/types.ts
semoal: i forgot to add the `RematchSelect` type, that's why it's copied

`		select: RematchSelect<TModels, RematchRootState<TModels>>	`

packages/select/src/types.ts
semoal: for nothing, on a first instance i've tried to use it in createModel in core, and extend the type here with selectors

packages/select/src/types.ts
semoal: could you give me an example? i'm not sure how to fill 

`selectors?: ModelSelectorsConfig<TState>`, for the model

packages/select/src/index.ts
semoal: like `createModel<RootModel, ExtraSelectorsModel<RootModel>>()({})`?

packages/select/package.json
```diff
@@ -24,7 +24,8 @@
     "src"
   ],
   "dependencies": {
-    "reselect": "^4.0.0"
+    "reselect": "^4.0.0",
+    "redux": "^4.0.0"
```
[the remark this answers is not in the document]
semoal: yes, redux should be in peerDeps

packages/select/src/types.ts
semoal: when select plugin it's inserted the store contains a new prop `store.select()`

```jsx
const selection = store.select(models => ({
  // todo: this isn't getting autocompleted
  total: models.cart.total,
}))

const mapState = (state: RootState) => ({
	settingsState: state.settings,
  loadingState: state.loading,
  playersState: state.players,
  // todo: 2n param props shouldn't be required
  ...selection(state, null)
})
```

packages/select/src/index.ts
tianzhich: I am not sure hot to extend native createModel for plugins. If we want the automation from selectors just like effects, seems the only way to work is yours. But if we have another plugin which also need model config. Are we create a helper `createModelWithXXX`? How could we use them together?

Or we can tell our plugin contributers to modify `createModel` in core and add these config definitions, if needed.

semoal:
@tianzhich let's see if we can give this pull request a push/sprint, at least some minimal features like slice it's already done. We should decide createModelSelectors if is interesting or we should extends createModel({}) to something else..

@pjamrozowicz If you have some free time your opinion or help will be really appreciated, this is the only pending task from your migration to ts :)

tianzhich:
@semoal yes, we can give a sprint.

And I used an overload for `createModel` just like what I did for `connect` function of `react-redux`, which was more elegant than `createModelSelectors`. So we can add extra plugin configs efficiently.

packages/select/src/types.ts
semoal: Sounds nice to me

haikyuu:
@semoal Is it safe to use rematch v2 before this PR is merged? I'm not using the select plugin?

semoal:
> @semoal Is it safe to use rematch v2 before this PR is merged? I'm not using the select plugin?

Yes absoutly, v2 won't change we're just improving/fixing select plugin development experience on typescript. But v2 is out of bugs, at least for now.

So feel free to use the next version. :)

haikyuu:
Thanks @semoal

semoal:
Probably should be a good idea to release this @tianzhich so select Users can start working and give US feedback of whats is pending?

tianzhich:
> Probably should be a good idea to release this @tianzhich so select Users can start working and give US feedback of whats is pending?

sure, we need to release this. Now only a few typing problems in tests. We need more feedbacks.

PR #833, opened 2020-11-24
feat: support optional payload parameter on reducer

loveky:
In my project, I came across the following scenario. 

I have a shared store that controls whether a global loading icon should show/hide. The global loading components supports an optional text.

Obviously, I call two reducers for each store field.  I'm thinking if we can do
```javascript
dispatch.globalLoading.show();  // only show loading icon

dispatch.globalLoading.show('Please wait...') // show loading icon with text
```
 
which is more simple and easy to use(in my opinion) 😃

### DIFF
--- a/examples/hooks-react-ts/src/models/dolphins.ts
+++ b/examples/hooks-react-ts/src/models/dolphins.ts
@@ -5,7 +5,7 @@ import { RootModel } from '.';
 export const dolphins = createModel<RootModel>()({
 	state: 0,
 	reducers: {
-		increment: (state) => state + 1,
+		increment: (state, payload: number = 1) => state + payload,
 	},
 	effects: (dispatch) => ({
 		async incrementAsync() {
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -337,7 +337,9 @@ export type ExtractRematchDispatcherFromReducer<
 > = TReducer extends () => any
 	? RematchDispatcher
 	: TReducer extends (state: TState) => TState
-	? RematchDispatcher
+	? Parameters<TReducer> extends [TState]
+		? RematchDispatcher
+		: RematchDispatcher<Parameters<TReducer>[1] | undefined>
 	: TReducer extends (state: TState, payload: infer TPayload) => TState
 	? RematchDispatcher<TPayload>
 	: never
@@ -350,6 +352,8 @@ export type ExtractRematchDispatcherFromReducer<
  */
 export type RematchDispatcher<TPayload = void> = [TPayload] extends [void]
 	? (() => Action<void>) & { isEffect: false }
+	: undefined extends TPayload
+	? ((payload?: TPayload) => Action<TPayload>) & { isEffect: false }
 	: ((payload: TPayload) => Action<TPayload>) & { isEffect: false }
 
 /** ************************ Effects Dispatcher ************************* */
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -339,7 +339,7 @@ export type ExtractRematchDispatcherFromReducer<
 	: TReducer extends (state: TState) => TState
 	? Parameters<TReducer> extends [TState]
 		? RematchDispatcher
-		: RematchDispatcher<Parameters<TReducer>[1] | undefined>
+		: RematchDispatcher<Parameters<TReducer>[1]>
 	: TReducer extends (state: TState, payload: infer TPayload) => TState
 	? RematchDispatcher<TPayload>
 	: never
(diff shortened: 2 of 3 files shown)


packages/core/src/types.ts
```diff
@@ -337,7 +337,9 @@ export type ExtractRematchDispatcherFromReducer<
 > = TReducer extends () => any
 	? RematchDispatcher
 	: TReducer extends (state: TState) => TState
-	? RematchDispatcher
+	? Parameters<TReducer> extends [TState]
+		? RematchDispatcher
+		: RematchDispatcher<Parameters<TReducer>[1] | undefined>
```
tianzhich, line 342: How about just `RematchDispatcher<Parameters<TReducer>[1]>` cause `Parameters<TReducer>[1]` include `undefined` in this branch.

tianzhich (changes requested):

examples/hooks-react-ts/src/models/dolphins.ts
```diff
@@ -5,7 +5,7 @@ import { RootModel } from '.';
 export const dolphins = createModel<RootModel>()({
 	state: 0,
 	reducers: {
-		increment: (state) => state + 1,
+		increment: (state, payload: number = 1) => state + payload,
```
tianzhich, line 8: Could we leave the `: number` annotation?

tianzhich:
Thanks for using rematch, we actually need to support optional payload. How about your opinion mate? @semoal

packages/core/src/types.ts
loveky: nice catch, PR updated

examples/hooks-react-ts/src/models/dolphins.ts
loveky: without `:number` annotation, the type of `payload` would be `any`. Because:

image

tianzhich: Yes exactly,  I also have some doubts here: TS should know the payload type by assigning a numeric value to it but it couldn't... 

Do you have any ideas bro?

tianzhich: I am not sure if it was an [issue](https://github.com/microsoft/TypeScript/issues/41668) with TS. 

And I also created a [playground](https://www.typescriptlang.org/play?ts=4.2.0-dev.20201124#code/KYDwDg9gTgLgBAE2AYwDYEMrDjAnmbAWQiVQCVgEBXZYKAZwB4AVAZRnRmwF450A7XAD44vAN4BYAFABIANoBrYLgBccejCgBLfgHMAumorVaUFu07Ah0gL7TpoSLEQoMWHPmzGadcxy6ifIIivAAU0jIalmps-sAANBFg6LioEOgIAPxqArjSAJSiIrGW9lLIEPwacMhYlsSkaqGF3CKMEWRwoFz8CPRwDcDklD4MjKxCibKs0kKhALYQapKy-OjzwNnqmjq6EVFcajOyWCZ09Ftkti0iKzJrG1sa2nr7cUcRp6P0RraBzUU4AsIPkcoJASs4HAsDAqFB+HAxAA6FGLOxSdHSCpVeC1YD1EhDADqWhgAAsIFQYN5TOxtMgYFpKj8gTc4O1ZGQpjJjnNFssIg9Nmpnrs3tE4McZF9TBdfhi2XchU8dq9ZAdgB8TiNZfKbP82cDQUFcBDpFCYXCEcjURB0ZipAB6R1wZKpdIIV2Ydb9HRwAAG9GAMFYEA25N2-pqAjg-Ag8AARtgdAAzOiUPj9fhUeZJqBY5nwFMQCCBPEE0jNUKQ7YSgCMUwtOvOAqkUKhQZDYeDZN2oQ1amzubo8S97oygTrhRr7ehwatta4AG5ze30VD7fkys6x2kJ8koD64H7A8HQ+He3oo8gY0nj-w06dM7Gc3mCzi4AnMGW6lxBqgSXJSlqWbKA6S0BkmSqKsawHOAG1XGUW0RVcOzPbsIz0ft3hfYcoFHN0909XgpxQttZznWF4UXYAV3IjdV03bcXUIj0vUPeZfQRU8uwvSNowRO9U3TT10CzV86DgAB3UkyS6cBUAg0kPAIII438KD32qL8AC8f3xP9CVQGDVzghDyKQhhW1nTtzx7Ps4KHPMCJSIjBwkqBJ2nVD20taiNTo2d1zgJipCAA) which you can have a look if you are interested.

semoal: There isn't support for partial type inference yet. There are some workarounds like currying https://github.com/microsoft/TypeScript/pull/26349, so probably this should stay like this until Typescript fixes this :)

semoal:
Looks good to me @loveky thanks for your contribution! Really appreciated

examples/hooks-react-ts/src/models/dolphins.ts
tianzhich: Yes, my issue marked as `design limitation`. I will merge this

tianzhich (approved):

PR #835, opened 2020-12-04
fix(loading): fix ts type when asNumber enabled

loveky:
### What's the problem?

The `loading` state value has a `boolean` type even when `asNumber` is enabled.

WX20201204-114827

### Why this happens?

image

### What I did to fix that?

Add an optional parameter to `ExtraModelsFromLoading` to let the developer specify whether `asNumber` is enabled.

### DIFF
--- a/docs/plugins/loading.md
+++ b/docs/plugins/loading.md
@@ -88,6 +88,15 @@ export const store = init<RootModel, FullModel>({
 	  plugins: [loadingPlugin()],
 })
 
+// or
+// type FullModel =  ExtraModelsFromLoading<RootModel, { asNumber: true }>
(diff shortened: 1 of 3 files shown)


loveky:
Result:

WX20201204-114913

tianzhich (commented):
Thank you loveky, nice gotcha. I missed it here.

And I add some changes to keep the same naming and construct here.

tianzhich:
check here https://github.com/loveky/rematch/pull/1

semoal:
Great guys looks good to me

semoal:
I'll release a new -next version in a few minutes

PR #847, opened 2020-12-20
fix: Improved overall bundle size

semoal:
Removing for-of, for-in iterators remove a lot of code


semoal:
> ## size-limit report 📦
> Path	Size
> ./packages/core/dist/core.umd.production.min.js	1.49 KB (-24.06% 🔽)

We decreased the size a 24% just removing for-of, and for-in :shipit: @tianzhich

packages/core/src/bag.ts
```diff
@@ -13,13 +13,11 @@ export default function createRematchBag<
 		models: createNamedModels(config.models),
 		reduxConfig: config.redux,
 		forEachPlugin(method, fn): void {
-			for (const plugin of config.plugins) {
+			config.plugins.forEach((plugin) => {
 				if (plugin[method]) {
-					// @ts-ignore
-					// temporary ignore, see: https://github.com/microsoft/TypeScript/issues/40429
-					fn(plugin[method])
+					fn(plugin[method]!)
```
semoal, line 18: @tianzhich should we move to this non-null assertion or keep the ts-ignore?

tianzhich: yes, we can use non-null assertion.

PR #851, opened 2020-12-26
fix: add missing variable for import statement

Bijiabo:
When I refer to the TypeScript sample code, I found there is a missing variable in `import { RootState, Dispatch } from './store'` statement. so...it is suggested to add it :)


semoal:
Lgtm thanksss for your contribution mate

PR #859, opened 2021-01-29
docs(issue-templ): add reproduce link section and use comment blocks for description.

tianzhich:
1. Add reproduce link section.
2. Use comment blocks for description.


tianzhich:
I updated the bug report issue template. Is that ok? @semoal

PR #860, opened 2021-01-29
chore: moved to yarn and improve dev experience

semoal:
- Moved to `yarn workspaces` to avoid lerna bootstrap (pretty slow, and yarn workspaces dependency handling is much better.
- Also I've run prettier to examples folder.

## Pending
- [ ] Integrate a verdaccio image to release @rematch packages and test them inside `all-plugins-react-ts` with a minimal suite that everything works correctly in the newer release.


semoal:
> ## size-limit report 📦
> Path	Size
> ./packages/core/dist/core.umd.production.min.js	1.43 KB (-0.88% 🔽)
> ./packages/immer/dist/immer.umd.production.min.js	138 B (0%)
> ./packages/loading/dist/loading.umd.production.min.js	596 B (-0.67% 🔽)
> ./packages/persist/dist/persist.umd.production.min.js	172 B (0%)
> ./packages/select/dist/select.umd.production.min.js	456 B (0%)
> ./packages/updated/dist/updated.umd.production.min.js	407 B (-0.74% 🔽)

We decreased even more the bundlesize updating deps 👯 @tianzhich

PR #867, opened 2021-02-19
fix: immer `whitelist` and `blacklist` conf no longer ignored

andrejpavlovic:
Specifying arrays of `whitelist` and/or `blacklist` models in `immerPlugin` constructor was being ignored.

I've made a fix and added test cases.


semoal:
LGTM super @andrejpavlovic thanks for the contribution mate, really appreciated

semoal:
I'll release this on tuesday, i want to fix some bugs

PR #869, opened 2021-02-22
fix: simplified config building

semoal:


semoal:
Though that this pr would decrease the size, but not. so byebye

Issue #872, opened 2021-03-01
bind here does nothing


objectisundefined:
https://github.com/rematch/rematch/blob/917b50eeeb73ad222068726060643a60a455295c/packages/core/src/rematchStore.ts#L54

should it be replaced by `rematchStore.addModel = rematchStore.addModel.bind(rematchStore)` or just replace `this` to `rematchStore` in function `addModel`?

https://github.com/rematch/rematch/blob/917b50eeeb73ad222068726060643a60a455295c/packages/core/src/rematchStore.ts#L43-L50

And there are no `addModel` test cases like assigning `rematchStore.addModel` to a a new function.

semoal:
Yes, looking at the code probably it should be like this:
```ts
		addModel(model: NamedModel<TModels>) {
			this.addModel.bind(rematchStore)
			validateModel(model)
			createModelReducer(bag, model)
			prepareModel(this, bag, model)
			this.replaceReducer(createRootReducer(bag))
			reduxStore.dispatch({ type: '@@redux/REPLACE' })
		},
	} as RematchStore<TModels, TExtraModels>
```

I don't understand the case of adding addModel to a new function. Contributions are welcome, if you want to submit this changes, we'll be great =)

PR #873, opened 2021-03-01
fix: context binding in addModel

objectisundefined:
for issue https://github.com/rematch/rematch/issues/872

### DIFF
--- a/packages/core/src/rematchStore.ts
+++ b/packages/core/src/rematchStore.ts
@@ -43,15 +43,13 @@ export default function createRematchStore<
 		addModel(model: NamedModel<TModels>) {
 			validateModel(model)
 			createModelReducer(bag, model)
-			prepareModel(this, bag, model)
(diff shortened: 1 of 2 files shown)


packages/core/test/init.test.ts
```diff
@@ -186,6 +186,23 @@ describe('init:', () => {
 			},
 		})
 	})
+
+	test('addModel works when it was used alone', () => {
```
semoal, line 190: move this test better to `store.test.ts` please, we already have there some tests related to storeModel =)

semoal (changes requested):

semoal (approved):

tianzhich:
Looks great! The `bind` does nothing here, change `this` to `rematchStore` would be better. Thanks for contributing @objectisundefined

tianzhich:
I merged, you can publish in your free time @semoal

PR #874, opened 2021-03-01
feat: introduced @rematch/typed-state plugin

semoal:
I think it's a cool addition to javascript projects that want to get some safe access to his stores.

### DIFF
--- a/packages/typed-state/src/index.ts
+++ b/packages/typed-state/src/index.ts
@@ -1,0 +1,106 @@
+import {
+	ModelEffects,
+	ModelEffectsCreator,
+	ModelReducers,
+	Models,
+	Plugin,
+} from '@rematch/core'
+import { Reducer } from 'redux'
+
+const cachedTypings: Record<string, any> = {}
+
+declare module '@rematch/core' {
+	export interface Model<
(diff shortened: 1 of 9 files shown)


semoal:
@tianzhich we you have time there's no hurry if you can take a look to typings, i think if we introduce together this plugin with select plugin won't work because typescript doesn't know merge multiple declare module '@rematch/core'

packages/typed-state/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+				)
+			}
+			return called
+		},
+	}
+}
+
+export default typedStatePlugin
+
+declare module '@rematch/core' {
```
tianzhich, line 99: Do you mean this type declaration won't work?

tianzhich: From the [TS docs](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#introduction):
> Any number of declarations can be merged; it’s not limited to just two declarations

semoal: This type declaration won't work. Loses the selectors type, like typescript isn't deep merging.
Screenshot 2021-03-02 at 09 48 31

PR #879, opened 2021-03-24
docs: fix immer plugin href

sunOpar:
## problem

If you in [docs](https://rematchjs.org/docs/), and then click plugin hyperlink in **Composable Plugins**. After that, your current location is `https://rematchjs.org/docs/plugins`. Finally, click the plugin list such as **Immer Plugin** hyperlink, you will get `Page Not Found`.


semoal:
Really appreciated the contribution mate =)

PR #892, opened 2021-05-13
fix: make models on init() Partial<T>

semoal:
https://github.com/rematch/rematch/pull/891/files#diff-1f8711ffcf4ef0b26aea69d164072937102d5109ecfd8790b6e879c7421e00d3

Since I've checked how it's the strategy with typescript and `addModel`, has sense to define a RootModel which is Partial<T> of a complete RootModel.

```ts
import { Models } from '@rematch/core'
import { players } from './players'
import { settings } from './settings'
import { cart } from './cart'

export interface RootModel extends Models<RootModel> {
	players: typeof players
	cart: typeof cart
	settings: typeof settings
}

export const models: Partial<RootModel> = { players, cart }
```

In this way we can define the complete typings but a FUTURE usage, but the current models isn't complete yet.

### DIFF
--- a/packages/core/src/bag.ts
+++ b/packages/core/src/bag.ts
@@ -30,9 +30,9 @@ export default function createRematchBag<
  */
 function createNamedModels<
 	TModels extends Models<TModels> = Record<string, any>
->(models: TModels): NamedModel<TModels>[] {
+>(models: TModels | Partial<TModels>): NamedModel<TModels>[] {
 	return Object.keys(models).map((modelName: string) => {
-		const model = createNamedModel(modelName, models[modelName])
+		const model = createNamedModel(modelName, (models as TModels)[modelName])
 		validateModel(model)
 		return model
 	})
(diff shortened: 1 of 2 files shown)


tianzhich (commented):
I am still not sure why we need to partialize the defined `rootModels`. If we don't use this model, we should also remove its typing? Could you explain more clearly mate?

tianzhich:
I seem to understand your opinion. Do you want to fix the dynamically added model's typing problem? I guess you want to define it's types before you add it dynamically. So there will no typing problem after it is added. Am I right?

semoal:
> I seem to understand your opinion. Do you want to fix the dynamically added model's typing problem? I guess you want to define it's types before you add it dynamically. So there will no typing problem after it is added. Am I right?

Exactly that

tianzhich (approved):
Well, I think that's well for dynamically adding features. Good job mate  👍

PR #901, opened 2021-05-25
fix: optional payload inference

tianzhich:
1. Simplify the `ExtractRematchDispatcherFromReducer` type alias.  
2. Make the Inference of the optional payload (and meta) correctly.


semoal:
Looks good mate, nice refactor :) 

I've created a file called `packages/core/test/ts_typings/dispatcher-typings.test.ts`, to create tests of our typings related to dispatch, basically consists:

- Reducers
	- 	With optional payload
	-  With required payload
- Effects
	- 	With optional payload
	-  With required payload
	- With optional payload and rootState
	- With required payload and required rootstate

and saw that effects with optional payload was not working properly, take a look to this pr mate: #902

PR #902, opened 2021-05-25
fix: optional payload inference inside effects

semoal:
Related to #901 

Introduces a new test to avoid regression with dispatcher system:
- Reducers
	- With optional payload
	- With required payload
- Effects
	- With optional payload
	- With required payload
	- With optional payload and rootState
	- With required payload and required rootstate


tianzhich (approved):
Now all the existing tests look great. We can add more tests or merge this first and append tests in the future.

Issue #894, opened 2021-05-16
Improve Typescript dispatch reducers inference


semoal:
```ts
import { createModel } from '@rematch/core'
import { delay } from './utils'
import type { RootModel } from '.'

export const dolphins = createModel<RootModel>()({
	state: 0,
	reducers: {
		increment: (state, payload = 1) => state + payload,
	}
})

```
**Describe the bug**

Actually the `payload = 1` is infered as any and when dispatching the reducer says that: `(property) increment: () => Action<void, void>`, doesn't expect a payload...

The same occurs if the payload it's optional

tianzhich:
I've fixed it if the payload is optional.

tianzhich:
Maybe it is [design limitation](https://github.com/microsoft/TypeScript/issues/41668).

semoal:
Close this, since will be fixed with 2.1.0 release

Issue #893, opened 2021-05-16
Typescript `connect()` fails about circular references


semoal:
**Describe the bug**

When updating Typescript from 4.1.x to 4.2.x, I've observed that `connect()` method exported from `react-redux` has a type error.

<img width="1154" alt="Screenshot 2021-05-16 at 7 25 52 PM" src="https://user-images.githubusercontent.com/22656541/118406589-af689a00-b67c-11eb-8758-9f028bff1f57.png">

**Expected behavior**

To not fail

**Additional context**

The possibly solution is to just use `any` instead of using RematchDispatch<RM>, I think this change doesn't break anything.

<img width="789" alt="Screenshot 2021-05-16 at 7 29 04 PM" src="https://user-images.githubusercontent.com/22656541/118406669-0ec6aa00-b67d-11eb-81e1-175936bf5f20.png">

tianzhich:
I'll pick this one and try to solve it this week

tianzhich:
I found the error cause, but not sure if this is an issue with typescript. 

Our `RematchDispatch`: 

```ts
export type RematchDispatch<
	TModels extends Models<TModels> = Record<string, any>
> = ReduxDispatch & ExtractRematchDispatchersFromModels<TModels>
```

If I removed the hybrid part of `ReduxDispatch`, the error gone.

tianzhich:
I tried to create a minimal reproduce but I failed. If some could help more appreciated.

semoal:
You have it in the count-react-ts file a repro.

tianzhich:
> You have it in the count-react-ts example a repro. Be sure to update the typescript version to 4.2.x

You are right, but I mean to create a typescript repro based on a minimum of our types, though I'm not sure if it is an issue with typescript. I searched circular reference error, but few related issues were found.

tbaliukynas:
I noticed that using `RematchRootState<RootModel>` and `RematchDispatch<RootModel>` directly instead of `RootState` and `Dispatch` defined in `store.ts` seems to fix the issue (at least in the `count-react-ts` example):
```diff
diff --git a/examples/count-react-ts/src/App.tsx b/examples/count-react-ts/src/App.tsx
index fb0774a..5b1eaa6 100644
--- a/examples/count-react-ts/src/App.tsx
+++ b/examples/count-react-ts/src/App.tsx
@@ -1,13 +1,14 @@
+import { RematchDispatch, RematchRootState } from '@rematch/core'
 import * as React from 'react'
[...]

Issue #870, opened 2021-02-25
On effects, this.reducer() is not correctly infered


semoal:
**Describe the bug**

With Rematch you can use `this` inside effects to dispatch data to reducers.

Basically this:

```tsx
export const players = createModel<RootModel>()({
	state: {
		players: [],
	},
	reducers: {
		SET_PLAYERS: (state, players) => {
			return {
				...state,
				players,
			}
		},
	},
	effects: () => ({
    async getPlayers() {
      const response = await fetch(
        'https://www.balldontlie.io/api/v1/players'
      )
      const { data } = await response.json()
      this.SET_PLAYERS(data)
    }
  })
})
```

is equivalent to this:

```tsx
export const players = createModel<RootModel>()({
	state: {
		players: [],
	},
	reducers: {
		SET_PLAYERS: (state, players) => {
			return {
				...state,
				players,
			}
		},
	},
	effects: (dispatch) => ({
    async getPlayers() {
      const response = await fetch(
        'https://www.balldontlie.io/api/v1/players'
      )
      const { data } = await response.json()
      dispatch.players.SET_PLAYERS(data)
    }
  })
})
```

**To Reproduce**

When using Typescript, this.SET_PLAYERS() won't get autocompleted because this isn't overriden by Rematch.

**Current work around**

```
      (this as typeof players).SET_PLAYERS(data)

````

**Expected behavior**

Should autocomplete reducers and effects of the current Model.

**Additional context**

- Version: 2.0.1

**Comment by maciekgrzybek** [comment]
I'm happy to take a look at this :)

**Comment by maciekgrzybek** [comment]
It's really tough to figure it out really... Do you have any ideas what might be causing that?

**Comment by semoal** [comment]
Actually this type:
```ts
export type ModelEffect<
	TModels extends Models<TModels> = Record<string, any>
> = (
	payload: Action['payload'],
	rootState: RematchRootState<TModels>,
	meta: Action['meta']
) => any
```

Is the current type for an effect, you will see that if you add the `this` property with anything else, you get autocomplete of the rootstate:

```ts
export type ModelEffect<
	TModels extends Models<TModels> = Record<string, any>
> = (
    this: RematchRootState<TModels>
	payload: Action['payload'],
	rootState: RematchRootState<TModels>,
	meta: Action['meta']
) => any
```

We must add something like: `ExtractReducers&Effects<TModel>` where we can pass a Model and extract his effects and reducers. 
 
[...]

PR #903, opened 2021-05-26
refactor: improve type system

tianzhich:
There are some fixes or improvements related to the rematch type system. If you have any issues you can checkout a branch from the `improve-typing` and create a PR and set `improve-typing` as merge target.

### Bug fixes
- [x] #893 - Typescript connect() fails about circular references
- [X] #894 - #902 Improve Typescript dispatch.reducer inference and effects
- [x] #870 - this.reducer isn't typed correctly, probably this one since it's complex to fix, we can go with any.
- [X] #892 - init() models also accepts Partial<RootModel>, this can be used in conjuntion with store.addModel() for dynamic adding models

### Improvements
- [x] Remove all the default generic type variable `Record<string, any>`. If needed, use `Record<string, unknown>` or `Record<string, never>` instead.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -246,10 +246,12 @@ export interface InitConfigRedux<TRootState = any> {
 	enhancers?: StoreEnhancer[]
 	middlewares?: Middleware[]
 	rootReducers?: ReducersMapObject<TRootState, Action>
-	combineReducers?: (
-		reducers: ReducersMapObject<TRootState, Action>
-	) => ReduxReducer<TRootState>
-	createStore?: StoreCreator
+	combineReducers?:
+		| ((
+				reducers: ReducersMapObject<TRootState, Action>
+		  ) => ReduxReducer<TRootState>)
+		| undefined
+	createStore?: StoreCreator | undefined
 	devtoolOptions?: DevtoolOptions
 }
 
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -88,9 +88,14 @@ export interface ModelEffects<
 	[key: string]: ModelEffect<TModels>
 }
 
+export type ModelEffectThisTyped = {
+	[key: string]: (payload?: any, meta?: any) => Action<any, any>
+}
+
 export type ModelEffect<
 	TModels extends Models<TModels> = Record<string, any>
 > = (
+	this: ModelEffectThisTyped,
 	payload: Action['payload'],
 	rootState: RematchRootState<TModels>,
 	meta: Action['meta']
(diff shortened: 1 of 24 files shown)


semoal:
> Remove all the default generic type variable Record<string, any>. If needed, use Record<string, unknown> or Record<string, never> instead.

This one, could end in a breaking change? Or this won't impact the end user? Also, what's the benefit? Just curiosity ^^

tianzhich:
No, this won't end in a breaking change and have no impact on the end-user. 

My main consideration is too much redundant code is added if using the default `Record<string, any>`. The `TModel` and `TModels` mast have value.

And the second point is `Record<string, any>` not safe as `Record<string, unknow>` of `Record<string, never>`. If you use `dispatch.notExistModel.doSomething` you get any error.

semoal:
Zhi Tian, mate, what do you think about this?


###  #870 - this.reducer isn't typed correctly, probably this one since it's complex to fix, we can go with any.

Possible solution:

```ts
export declare type ModelEffect<TModels extends Models<TModels> = Record<string, any>> = (
	  this: any,
	  payload: Action['payload'],
	  rootState: RematchRootState<TModels>,
	  meta: Action['meta']
) => any;
```

Since we don't know yet and Typescript it's pretty limited to this use cases at least for the moment, I think a good solution is to type this to any to not see any error in this use case.


### #893 - Typescript connect() fails about circular references

Here the same, I would use here at `any` since doesn't affect us on anything.


I'm already writing the Chapter 10 of the book which includes a complete rewrite of the application to Typescript and practically everything worked out of the box nicely, only the `this.reducer` use case and the optional payload argument, which it's already fixed here :)

tianzhich:
No problem mate, for #870 we may type this to `this: { [key: string]: (payload?: any, meta?: any) => Action<any, any> }` just like the v1, it's more specific than `any`. For the #893 , type as `any` or just upgrade typescript later both ok for me.

Good job mate! So I can review the next chapters. Looking forward to your new book :)

packages/core/src/types.ts
```diff
@@ -246,10 +246,12 @@ export interface InitConfigRedux<TRootState = any> {
...
-	combineReducers?: (
-		reducers: ReducersMapObject<TRootState, Action>
-	) => ReduxReducer<TRootState>
-	createStore?: StoreCreator
+	combineReducers?:
+		| ((
+				reducers: ReducersMapObject<TRootState, Action>
+		  ) => ReduxReducer<TRootState>)
+		| undefined
+	createStore?: StoreCreator | undefined
```
tianzhich, line 254: Why we need the explicit `any`?

semoal: Typrescript 4.4.x fails about error values used in catch methods which are not typed as any or unknown.

semoal: Ah didn't read you correctly. Ts 4.4 fails about not compatible X | UNDEFINED to X. 


Probably it's a regression I'll take a look before merging into main

tianzhich:
We are ready for this thred.

semoal (approved):

PR #904, opened 2021-06-13
refactor: select typings without Record<string,any>

semoal:
Co-authored-by: tianzhich <zhi.tian@yahoo.com>


tianzhich (approved):

tianzhich:
@semoal mate, do you know why size testing failed?

semoal:
Just because it only runs on main

PR #907, opened 2021-06-30
fix: loading plugin returns chain of promises effect

### LINKED ISSUE #906: Unhandled promise rejection with the Loading plugin
A Warning Unhandled promise rejection is thrown when having a Promise.reject using async function with the Loading plugin

**Reproduce Link**

Base : https://codesandbox.io/s/rematch-todos-demo-forked-xlq9k?file=/models.js

**To Reproduce**

1. In the method asyncRemove add at the top a throw (eg: throw "error"; )
2. Where the asyncRemove is called surround it with a try catch
2. Open the debugging console
3. In the web view click on the button async delete

**Expected behavior**

In the loading wrapper it should catch the error and rethrow ii

**Platform**

Web
Mobile (React-Native)

SpiriaJWF:
Return the chain promise to propagate the throw error and continue the chaining normally. 
Fix the Unhandled Promise Rejection

Closes #906

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+    setupFiles: ['<rootDir>/testsetup.js']
+};
--- a/packages/loading/src/index.ts
+++ b/packages/loading/src/index.ts
@@ -233,7 +233,7 @@ export default <
 						// check if result is a promise
 						if (effectResult?.then) {
 							// hide loading when promise finishes either with success or error
-							effectResult
+							return effectResult
 								.then((r: any) => {
 									rematch.dispatch[loadingModelName].hide({ name, action })
 									return r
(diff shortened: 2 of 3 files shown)


semoal:
Looks great to me and has a lot of sense, could you add a test for this?
I have a plan to introduce a big feature inside loading plugin and I don't want to break any suit case

SpiriaJWF:
Well I don't know if it would add anything. You already have tests that cover the propagation of the effect error. The new code simply prevent the warning.

SpiriaJWF:
I guess we could put this in the describe function?

`
process.on('unhandledRejection', (err) => {
    fail(err);
});
`

semoal:
Thanks for the contribution mate, I'll release this with 2.1.0 in aprox 1 week more or less it will be released. :)

PR #909, opened 2021-07-04
chore: fix updated example error

ricosmall:


semoal:
Hello, thanks for the contributon but next branch is "old". We are using "main" branch. Could you check if this issue exists also in main branch?

semoal:
Hello, thanks for the contributon but next branch is "old". We are using "main" branch. Could you check if this issue exists also in main branch?

PR #911, opened 2021-07-06
docs: introduced Rematch book promotion

semoal:


website/src/pages/styles.module.css
```diff
@@ -289,6 +334,10 @@
   margin-left: 10%;
 }
 
+.techImage {
```
semoal, line 337: remove this

PR #913, opened 2021-07-06
fix: reducers and effects with same name are correctly typed 4.3.X

### LINKED ISSUE #912: Upgraded 4.1.2 to 4.3.X of TypeScript and never is returned when calling effect-reducer with same name
**Describe the bug**

Uupgraded from typescript 4.1.2 to 4.3.2 using Rematch. These rematch packages are used:

"@rematch/core": "2.0.1"
"@rematch/select": "3.0.1"

and I faced a typescript error: Type never has no call signatures. I looked into the 3rd parties' code and found that RematchDispatch gives never type when I have a reducer and an effect in a particular model with the same name. For example:

**Reproduce Link**

[codesandbox.io/s/interesting-sky-wnt9p?file=/src/Count.tsx](codesandbox.io/s/interesting-sky-wnt9p?file=/src/Count.tsx)

**Expected behavior**

To don't throw `never`

**Possible workaround** 

Meanwhile the typing issue is fixed, you can name differently the effect and the reducer

Og issue: https://stackoverflow.com/questions/68253239/why-does-rematchdispatch-give-never-type-when-there-is-a-reducer-and-an-effect

semoal:
Resolves #912 

Before 4.3 TypeScript accepted "duplicated" types but now is more stricter, so we just merge them giving priority always to reducers because is the entrypoint (will be the first executed if the same is duplicated)

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -21,6 +21,22 @@ import {
 } from 'redux'
 
 /**
+ * Utility type taken by type-fest repository
+ * https://github.com/sindresorhus/type-fest/blob/main/source/merge-exclusive.d.ts
+ * Merges Exclusively two types into one
+ * Used to fix this https://github.com/rematch/rematch/issues/912
+ */
+type Without<FirstType, SecondType> = {
+	[KeyType in Exclude<keyof FirstType, keyof SecondType>]: never
+}
+export type MergeExclusive<FirstType, SecondType> =
+	| FirstType
+	| SecondType extends object
+	?
+			| (Without<FirstType, SecondType> & SecondType)
+			| (Without<SecondType, FirstType> & FirstType)
+	: FirstType | SecondType
+/**
  * Custom Action interface, adds an additional field - `payload`.
  *
  * Strings (instead of Symbols) are used as the type for `type` field inherited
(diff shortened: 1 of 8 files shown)


tianzhich (approved):
Nice utility type. Good job mate. Do we need to add a notice about this case: if the same name used in both reducer and effects, the reducer will be executed first, and then the effects:

```ts
function createEffectsMiddleware<
	TModels extends Models<TModels>,
	TExtraModels extends Models<TModels>
>(bag: RematchBag<TModels, TExtraModels>): Middleware {
	return (store) => (next) => (action: Action): any => {
		if (action.type in bag.effects) {
			// first run reducer action if exists
			next(action)

			// then run the effect and return its result
			return (bag.effects as any)[action.type](
				action.payload,
				store.getState(),
				action.meta
			)
		}

		return next(action)
	}
}
```

I can't figure out this executing order until I read the above source code.

tianzhich:
Or we should remind our users not to use this way which I think is confusing. In my opinion, we can use a more semantic name on effects if it conflicts with the reducer.

semoal:
> Or we should remind our users not to use this way which I think is confusing. In my opinion, we can use a more semantic name on effects if it conflicts with the reducer.

Technically, our recommendation is to use Uppercase for reducers and effects with camelcase in that way you can differentiate easily on Redux Devtools. But... as we accept reducers and effects with the same name, we have to accept that on typings too

semoal:
> Nice utility type. Good job mate. Do we need to add a notice about this case: if the same name used in both reducer and effects, the reducer will be executed first, and then the effects:
> 
> ```ts
> function createEffectsMiddleware<
> 	TModels extends Models<TModels>,
> 	TExtraModels extends Models<TModels>
> >(bag: RematchBag<TModels, TExtraModels>): Middleware {
> 	return (store) => (next) => (action: Action): any => {
> 		if (action.type in bag.effects) {
> 			// first run reducer action if exists
> 			next(action)
> 
> 			// then run the effect and return its result
> 			return (bag.effects as any)[action.type](
> 				action.payload,
> 				store.getState(),
> 				action.meta
> 			)
> 		}
> 
> 		return next(action)
> 	}
> }
> ```
> 
> I can't figure out this executing order until I read the above source code.

oh, i thought it was the inverse... let me check

semoal:
@tianzhich Just introduced a new test to check the order of execution and don't introduce any regression there, also i replaced the order of the merge exclusive to give priority to reducers type.

Where the scenario of: 

```tsx
state: {
  count: 0,
},
reducers: {
  add(state, payload: string) {
    return state
  },
},
effects: () => ({
  add(payload: number) {
  },
}),
```

store.dispatch.add() will type the payload as string

tianzhich:
well, now it's no problem for me.

PR #918, opened 2021-07-15
docs: highlight correct line

frencojobs:
The documentation is looking pretty good. I just think this is the line you actually meant to highlight.


semoal:
Yep, totally missed that! Thanks Frenco for the contribution to this repo and of course shiki ;)

PR #922, opened 2021-08-11
feat: @rematch/loading with type: 'full' returns success/error/loading

### LINKED ISSUE #877: Suggestion: Error/success handling plugin
I love the way the  `@rematch/loading` plugin keeps my app's loading state all in one place. Something that stuck out to me as odd was that there's no similar plugin for handling error/success state. So I can now avoid writing a `{ loading, error, success }` object for every reducer, but `error` and `success` are still required if I want to perform some action after the effect finishes, so the returns are diminished. It would be amazing to have a plugin (or plugins) for these as well.

I understand there's a pretty robust plugin API for Rematch, so I'll probably end up creating this myself if it doesn't exist already, but I'm wondering if someone else has already built one and is willing to share.

semoal:
Finally found some time to work on this cool feature.

This introduces a little breaking change (which is not, because under the hood adapts the configuration to match the newest one if users are using the old one)

Basically now @rematch/loading plugin accepts a `type` property (By default is type: boolean)

This property accepts: `'number'`, `'boolean'`, `'full'`

### { type: "full" }

Basically returns `{ success: false, loading: true, error: false }`, if the effect promise returns an error the error response is added to error.

Works fine with Typescript as far as I could check.

# Pending

- [x] Decide if 'detailed' is a good name
- [x] Documentation
- [x] More testing
- [x] Reduce bundle size a bit (there're some optimisations) 

Will resolve #877

### DIFF
--- a/examples/all-plugins-react-ts/tsconfig.json
+++ b/examples/all-plugins-react-ts/tsconfig.json
@@ -17,7 +17,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react",
+    "jsx": "react-jsx",
     "noFallthroughCasesInSwitch": true
   },
   "include": [
--- a/packages/loading/src/index.ts
+++ b/packages/loading/src/index.ts
@@ -11,74 +11,116 @@ export interface LoadingConfig {
 	name?: string
 	whitelist?: string[]
 	blacklist?: string[]
+	type?: 'number' | 'boolean' | 'detailed'
+	// @deprecated Use type: 'number' instead
 	asNumber?: boolean
 }
 
-export type LoadingAsNumberConfig = Required<Pick<LoadingConfig, 'asNumber'>>
-
-interface LoadingState<
+interface LoadingStateV2<
 	TModels extends Models<TModels>,
-	AsNumber extends boolean = false
+	WhichType extends 'number' | 'boolean' | 'detailed'
 > {
-	global: AsNumber extends true ? number : boolean
+	global: WhichType extends 'number'
+		? number
+		: WhichType extends 'detailed'
+		? DetailedPayload
+		: boolean
 	models: {
-		[modelName in keyof TModels]: AsNumber extends true ? number : boolean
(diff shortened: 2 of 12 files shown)


tianzhich:
How about 'detail' or 'details'?

examples/all-plugins-react-ts/tsconfig.json
```diff
@@ -17,7 +17,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react",
+    "jsx": "react-jsx",
```
tianzhich, line 20: Why add `-jsx` suffix?

packages/loading/src/index.ts
```diff
@@ -11,91 +11,134 @@ export interface LoadingConfig {
...
 	TModels extends Models<TModels>,
-	AsNumber extends boolean = false
+	WhichType extends 'number' | 'boolean' | 'detailed'
 > {
-	global: AsNumber extends true ? number : boolean
+	global: WhichType extends 'number'
+		? number
+		: WhichType extends 'detailed'
+		? DetailedPayload
+		: boolean
```
tianzhich, line 27: I'll write a generic type to handle this which duplicates many times.

packages/loading/src/index.ts
```diff
@@ -11,91 +11,134 @@ export interface LoadingConfig {
...
 }
 
-type Converter = ((cnt: number) => number) | ((cnt: number) => boolean)
+type Converter = (
+	cnt: number,
+	detailedPayload?: DetailedPayload
+) => number | boolean | DetailedPayload
 
-interface LoadingModel<
+interface LoadingModelV2<
```
tianzhich, line 78: How about using `LoaidngModel` directly?

tianzhich (changes requested):

tianzhich:
plz check my commits and comments @semoal.

examples/all-plugins-react-ts/tsconfig.json
semoal: React 17 requires react-jsx on tsconfig

PR #928, opened 2021-09-13
fix: immer behaves unexpectedly when model state is a nullable primitive value

liuqiang1357:
By the way, why we treat primitive different from object value? I think the default behavior is ok.

### DIFF
--- a/packages/immer/src/index.ts
+++ b/packages/immer/src/index.ts
@@ -10,13 +10,9 @@ export type ImmerPluginConfig = {
 function wrapReducerWithImmer(reducer: Redux.Reducer) {
 	// reducer must return value because literal don't support immer
 	return (state: any, payload: any): any =>
-		typeof state === 'object'
-			? produce(state, (draft: any) => {
-					const next = reducer(draft, payload)
-					if (typeof next === 'object') return next
-					return undefined
-			  })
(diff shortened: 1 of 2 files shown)


packages/immer/src/index.ts
```diff
@@ -10,13 +10,9 @@ export type ImmerPluginConfig = {
 function wrapReducerWithImmer(reducer: Redux.Reducer) {
 	// reducer must return value because literal don't support immer
```
tianzhich, line 11: Is this mean reducer must return object value because immer don't support primitive values? @semoal

tianzhich (approved):
Yes, It definitely a rematch bug. Any primitive unions object type both have this problem. And it's acceptable to use null with any other primitives. What's your opinion? @semoal 

Thanks for contributing.

packages/immer/src/index.ts
tianzhich: If it means this way, we may need to remove this comment. Now immer support returns any value [except the `undefined`](https://immerjs.github.io/immer/return/#producing-undefined-using-nothing).

semoal: yep, lgtm just remove that line and we're ready to ship

tianzhich (approved):

Issue #936, opened 2021-09-28
Typescript error occurred after upgrading the project to version 2.0


liguowei01:
There is no type defined in the original code. The type must now be defined after the upgrade. If you can't use "any", it will increase a lot of work. Is there any way to solve it?
![image](https://user-images.githubusercontent.com/47907319/135064503-595f2f81-c031-4ae3-88fa-56bb2d259842.png)

![image](https://user-images.githubusercontent.com/47907319/135065569-49c51290-cb99-489f-a2c2-a5c447efe041.png)

semoal:
Imho, you should type your payloads, if you don't know what is the `payload` attribute, just use Record<string, any> or even use `//@ts-expect-error` until you can type the `payload` to anything more feasible to reality.

Type in your state:

```ts
type PlayersState = {
	players: Record<string, any>
}

...
	state: {
		players: [],
	} as PlayersState,

..
SET_PLAYERS: (state, players: Record<string, any>) => {
```

liguowei01:
Well, thank you

tianzhich:
This might a bug. I'll look into it.

PR #937, opened 2021-09-30
fix(types): improve the accuracy of dispatcher inference

tianzhich:
- [ ] fix: dispatcher has no parameters if `payload` defined as `any` type #936 
- [ ] fix: improve the accuracy of dispatcher inference, for example to differentiate the optional argument with user-defined `undefined` type.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -21,6 +21,42 @@ import {
 } from 'redux'
 
 /**
+ * Utility type used to check if the parameters in reducer/effects is optional
+ * Differentiate `[payload?: number]` from `[payload: number | undefined]`
+ * to improve ts experience
+ */
+type CheckIfParameterOptional<P> = P extends [unknown, ...unknown[]]
+	? false
+	: true
+
+/**
+ * Utility type used to extract the whole payload/meta parameter type
+ * from reducer/effect parameters
+ * For example, extract `[meta?: string]`
+ * from `[payload: number, meta?: string]`
+ */
+type ExtractParameter<
+	P extends unknown[],
+	V extends 'payload' | 'meta'
+> = P extends [p?: infer TPayload, m?: infer TMeta, ...args: unknown[]]
+	? [TPayload] extends undefined
+		? never
+		: [TMeta] extends [undefined]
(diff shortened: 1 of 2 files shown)


packages/core/src/types.ts
```diff
@@ -395,12 +433,12 @@ export type ExtractRematchDispatcherFromReducer<TState, TReducer> =
  * Otherwise, it describes dispatcher which accepts one argument (payload)
  * and returns an action.
  */
-export type RematchDispatcher<TPayload = void, TMeta = void> = [
+export type RematchDispatcher<TPayload = never, TMeta = never> = [
 	TPayload,
 	TMeta
-] extends [void, void]
+] extends [never, never]
```
tianzhich, line 439: Change `void` to `never` for better constraint.

```ts
type Foo = [any] extends [void] ? 1 : 0 // 1
type Bar = [any] extends [never] ? 1 : 0 // 0
```

packages/core/test/ts_typings/dispatcher-typings.test.ts
```diff
@@ -62,6 +62,50 @@ describe('Dispatcher typings', () => {
...
+				myModel: typeof model
+			}
+
+			const store = init({ models: { myModel: model } })
+
+			const { dispatch } = store
+			dispatch.myModel.inc(4)
+			dispatch.myModel.inc('4')
+			// @ts-expect-error
+			dispatch.myModel.inc()
```
tianzhich, line 86: mate @semoal we need to differentiate the `payload?: any` from `payload: any`. Or better, `payload?: number` from `payload: number | undefined`. In our current inference, we treat them the same way, users can use an optional payload. 
There wouldn't be any run-time errors thrown, but it's inconsistent with TypeScript bahevior.

tianzhich:
The building process blocked by rollup cause some TS errors(these errors didn't appear in my local with ts@4.4.2), can you have a look? @semoal 

I'll formate the `types` file and also add/amend the documents there, in the week after next. Next week is China's National Day.

tianzhich:
Could you create a reproduce repo or sandbox? The codes you posted above works in our tests.

PR #969, opened 2022-04-14
fix: circular reference in dispatch

robertaistleitner:
When reusing an existing model in another model that includes additional models (aka. extending it) there is a circular reference detected: `Type of property 'base' circularly references itself in mapped type 'ExtractRematchDispatchersFromModels<ExtendedModel>'.`

The new test introduced will fail at the moment which is intended, but a solution in `types.ts` is provided but the implications of the change are not 100% clear to me.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -352,9 +352,14 @@ export type RematchDispatch<TModels extends Models<TModels>> = ReduxDispatch &
 export type ExtractRematchDispatchersFromModels<
 	TModels extends Models<TModels>
 > = {
+	// this causes the "Type of property 'base' circularly references itself in mapped type 'ExtractRematchDispatchersFromModels<ExtendedModel>'" typescript error when reusing models
 	[modelKey in keyof TModels]: TModels[modelKey] extends Model<TModels>
 		? ModelDispatcher<TModels[modelKey], TModels>
 		: never
+
+	// this modification also fixes the broken circular dependency issue:
+	// QUESTION: why is the conditional type check even needed in the first place?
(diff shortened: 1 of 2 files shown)


robertaistleitner:
@semoal is there a chance you'll have a look on the MR soon - otherwise I'll do a workaround for our project temporarily.

semoal:
@tianzhich probably we should take a look at this here https://github.com/rematch/rematch/pull/960 what do you think?

semoal:
@robertaistleitner once @tianzhich resolves my question, we ship a new version with this fixed, meanwhile if you want to use patch-package and I ping you when we release a new version

tianzhich:
ok i'll take a look this weekend.

robertaistleitner:
@tianzhich may I ask for an update here?

robertaistleitner:
@tianzhich may I ask for an update here?

froston:
Hi, I'm having the same issue.
Even using the [example for typescript in the documentation](https://rematchjs.org/docs/getting-started/typescript/) I'm getting these errors. Is the library still maintained?

```typescript
import { createModel } from '@rematch/core';
import type { RootModel } from './';

export const count = createModel<RootModel>()({
  state: 0,
  reducers: {
    increment(state, payload: number) {
      return state + payload;
    },
  },
  effects: (dispatch) => ({
    incrementAsync(payload: number, state) {
      dispatch.count.increment(payload);
      return state.count;
    },
  }),
});
```

`Line 14: Type of property 'count' circularly references itself in mapped type '{ [x: string]: any; form: TypeFormState; options: TypeOptionsState; portal: TypePortalState; count: any; }`

`Line 4: 'count' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.`

